PR #7, opened 2017-06-01
Wiki branch for wiki-related content

Hjdskes:
The Wiki cannot host data such as images itself, so we need to host them somewhere.


Hjdskes:
Whoops, seems like you can't PR into a new branch. I'll just push this directly, then.

PR #9, opened 2017-06-08
Add initial GtkApplication subclass

Hjdskes:
A Gtk.Application subclass to handle the application's initialization and integration with the GNOME stack. It implements the do_startup and do_activate methods and is responsible for the application's menus, icons, title and lifetime.

PR #10, opened 2017-06-10
Update ratbagd bindings

Hjdskes:
The bindings weren't up to date, some properties, methods and capabilities were missing.

In the process, I made the bindings' classes inherit from GObject, so that the Piper code can connect to its signals and watch for property changes. The details are in the commit message of 4618ad2.

I'm not sure if there is anything missing (I modeled things off of the DBus interface as specified in [ratbagd's README](https://github.com/libratbag/ratbagd#the-dbus-interface)). I think that's fine; I'll add or adapt (or fix, it's currently untested) things when I find them to be missing while working on Piper.

One thing I did notice is that every `Ratbagd*` class instantiates a new DBus connection by inheriting from `_RatbagdDBus`. It seems unnecessary to me to have one DBus connection for each profile, resolution, button, et cetera. Is this design intentional or is this something we can/should improve?

I would like to add (at least some) unit tests, especially for the bindings. Is there a way to create a fake device or something like that for testing purposes?

I'll rebase #9 on top of this and get started on a skeleton window.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -21,8 +21,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 # DEALINGS IN THE SOFTWARE.
 
-from gi.repository import GLib
-from gi.repository import Gio
+from gi.repository import Gio, GLib, GObject
 
 class RatbagdDBusUnavailable(BaseException):
     """
@@ -30,8 +29,10 @@ class RatbagdDBusUnavailable(BaseException):
     """
     pass
 
-class _RatbagdDBus(object):
+class _RatbagdDBus(GObject.GObject):
     def __init__(self, interface, object_path):
+        GObject.GObject.__init__(self)
+
         self._dbus = Gio.bus_get_sync(Gio.BusType.SYSTEM, None)
         if self._dbus is None:
             raise RatbagdDBusUnavailable()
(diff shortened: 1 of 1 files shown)


Hjdskes:
Also, shouldn't the dbus methods be used in the getters directly to ensure we always return an up-to-date value?

piper/ratbagd.py
```diff
@@ -158,46 +176,45 @@ def active_profile(self):
...
-        return self._caps
+        return self.dbus_call("GetProfileByIndex", "u", index)
 
     def __eq__(self, other):
         return other and self._objpath == other._objpath
 
 class RatbagdProfile(_RatbagdDBus):
     """
-    Represents a ratbagd profile
+    Represents a ratbagd profile.
```
whot: fwiw, this is a english vs everything else thing - in english only proper sentences are followed by a full stop. In this case it wouldn't really count, a missing full stop would be appropriate (too). Just FYI

whot:
regarding the getters: we currently assume that we're the only ones accessing the device so the device should not change behind our back. This may change in the future, but for now it's good enough.

whot:
afaict this depends on https://github.com/libratbag/ratbagd/issues/29, I'll hold this off until we have the ratbagd side of things working for the capabilities (should be an easy patch).

piper/ratbagd.py
Hjdskes: The others had a full stop so I simply made it consistent, but it's rather nitpicky I agree :stuck_out_tongue:

Hjdskes:
^ Fixed in https://github.com/libratbag/ratbagd/pull/30

@whot What about this?

> One thing I did notice is that every `Ratbagd*` class instantiates a new DBus connection by inheriting from `_RatbagdDBus`. It seems unnecessary to me to have one DBus connection for each profile, resolution, button, et cetera. Is this design intentional or is this something we can/should improve?

whot:
oh, forgot about that, sorry. can't remember 100% but I think it's a side-effect of the quick and dirty design that the bindings had.

libratbag has a test device setup, but I'm not sure how to use that properly from ratbagd, especially because I don't think we compile it into the final library. but https://github.com/libratbag/libratbag/issues/179 should make that easier.

Hjdskes:
I opened #11 to track the DBus issue.

> libratbag has a test device setup, but I'm not sure how to use that properly from ratbagd, especially because I don't think we compile it into the final library. but libratbag/libratbag#179 should make that easier.

I'll look into this when the merge has been done. I think that that solves all points raised. Ready to merge, IMO.

PR #13, opened 2017-06-12
ratbagd: add missing commas

Hjdskes:
That's what you get for not testing...


Hjdskes:
It'd be nice to get write access so I can review and push these simple changes myself. Or directly push them to master.

PR #18, opened 2017-06-30
Conform to PEP8 style guide

Hjdskes:
No functional changes, just formatting. Style options are debatable; I don't really have any preference but it's good to have a standard for the code IMO. I went for PEP8 because it's the most used one. Checking for conformance should be done using the `flake8` tool and the configuration given in this commit, i.e. PEP8 specifies docstrings to be of 72 characters wide but since it passes flake8 it's ok.

E402 and E501 are ignored because we don't care (that much) about line length, and GObject Introspection requires us to import things as we currently do.

piper/piper.py has not been fixed as it is going to be removed eventually, when all its relevant bits have found a place in the new code.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+ignore = E402,E501
+exclude = .git,__pycache__,build,data,piper/piper.py
--- a/piper.in
+++ b/piper.in
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
-import sys
-import signal
-import os
-import locale
 import gettext
+import gi
+import locale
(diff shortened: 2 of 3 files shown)


Hjdskes:
> Would it make sense to exclude piper/piper.py in .flake8 then?

I suppose it would, until it's removed entirely. I will add this.

> Otherwise looks good, and I'll setup a circle-ci test that enforces this for PR :)

Thanks, I was going to discuss CI with @whot later but this solves that :)

Hjdskes:
`piper/piper.py` is now ignored.

PR #19, opened 2017-06-30
Update README

Hjdskes:
It already contains some dependencies that will be required when the MouseMap widget is merged, but I'll open a PR for that soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,57 +1,66 @@
-Piper
+Piper [![CircleCI](https://circleci.com/gh/libratbag/piper.svg?style=svg&circle-token=7082ad7a7fea706fff66f1547649dca32e446cb0)](https://circleci.com/gh/libratbag/piper)
 =====
 
-GTK application to configure gaming mice. Piper uses libratbag via ratbagd.
+Piper is a GTK+ application to configure gaming mice, using libratbag via
+ratbagd. For the design mockups, see the [Redesign
+Wiki](https://github.com/libratbag/piper/wiki/Piper-Redesign).
 
-Dependencies
-============
+In order to run Piper, `ratbagd` has to be running (without it, you'll get to
+see a pretty mouse trap). To see how, see [its
+README](https://github.com/libratbag/libratbag/blob/master/README.md#running-ratbagd-as-dbus-activated-systemd-service).
 
-Piper uses the ratbagd Python bindings. It doesn't require ratbagd itself to
-be running, but well, it'll be a little useless without ratbagd.
+Installing Piper
+================
 
-Sources
-=======
+Piper uses the [meson build system](http://mesonbuild.com/) which in turn uses
+[ninja](https://ninja-build.org/) to build and install itself. Run the following
+commands to clone Piper and initialize the build:
 
-https://github.com/libratbag/piper
+```
+$ git clone https://github.com/libratbag/piper.git
+$ cd piper
+$ meson builddir --prefix=/usr/
+```
 
-https://github.com/libratbag/ratbagd
-https://github.com/libratbag/libratbag
+To build or re-build after code-changes, run:
+
+```
+$ ninja -C builddir
+$ sudo ninja -C builddir install
+```
+
+Note: `builddir` is the build output directory and can be changed to any other
+directory name.
 
 Contributing
 ============
 
-Yes please. As you can see, years of design experience have gone into the
-user interface and it is clearly the pinnacle of what is possible.
-Feel free to send patches or pull requests or submit improvements for
-the general design of the app.
+Yes please. It's best to contact us first to see what you could do. Note that
+the devices displayed by Piper come from libratbag.
 
-Note that the devices displayed by Piper come from libratbag.
+Piper tries to conform to Python's PEP8 style guide. To verify your code before
+opening a PR, please install `flake8` and run the following commands to install
+its pre-commit hook:
 
-Screenshots
-===========
+```
+$ flake8 --install-hook git
+$ git config --bool flake8.strict true
+```
 
-![Piper screenshot](http://imgur.com/ELLCiSE.png)
+Source
+======
 
-Build Status
-============
+`git clone https://github.com/libratbag/piper.git`
 
-[![CircleCI](https://circleci.com/gh/libratbag/piper.svg?style=svg&circle-token=7082ad7a7fea706fff66f1547649dca32e446cb0)](https://circleci.com/gh/libratbag/piper)
+Bugs
+====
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,36 +1,78 @@
...
 
-Piper uses the ratbagd Python bindings. It doesn't require ratbagd itself to
-be running, but well, it'll be a little useless without ratbagd.
+* python3
+* gobject-introspection
+* gtk3
+* python-gobject
+* python-cairo
+* python-lxml
+* libratbag
```
bentiss: IIRC, @whot doesn't like having a full list of dependencies written in the README as it tends to be out of sync with the tree soon enough.

README.md
```diff
@@ -1,36 +1,78 @@
 Piper
 =====
 
-GTK application to configure gaming mice. Piper uses libratbag via ratbagd.
+Piper is a GTK+ application to configure gaming mice, using libratbag via
+ratbagd.
+
+Piper is currently being redesigned. For the mockups, see the [Redesign
+Wiki][wiki-redesign].
```
bentiss: Not sure this sentence should be turned like that.

How about?:
For the design mockups, see the [Redesign Wiki][wiki-redesign].

README.md
```diff
@@ -1,36 +1,78 @@
-Piper
+Piper [![CircleCI](https://circleci.com/gh/libratbag/piper.svg?style=svg&circle-token=7082ad7a7fea706fff66f1547649dca32e446cb0)](https://circleci.com/gh/libratbag/piper)
```
bentiss: This is probably a good idea to have the badge here.

Please rebase on top of master to remove the ugly "Build Status" entry I added.

README.md
```diff
@@ -1,36 +1,78 @@
...
 
-Yes please. As you can see, years of design experience have gone into the
-user interface and it is clearly the pinnacle of what is possible.
-Feel free to send patches or pull requests or submit improvements for
-the general design of the app.
+Yes please. Note, however, that Piper is currently being redesigned as a Google
+Summer of Code project and as such, things will change significantly. As such,
+it's best to contact us first to see what you could do. Otherwise, hang back and
+the work might be done for you! Current status can be seen [here][kanban], the
+design mockups can be seen on the [Wiki][wiki-redesign].
```
bentiss: I am not sure about this one too (@whot?).

My concerns are that yes, we should let know that there is an ongoing redesign progress, but writing this down in the README seems overkill, especially when we might forget to change this in the future.

How about a simple header on the github repository page (which I just edited)?

README.md
```diff
@@ -1,36 +1,78 @@
...
-===========
+Piper tries to conform to Python's PEP8 style guide. To verify your code before
+opening a PR, please install `flake8` and run the following commands to install
+its pre-commit hook:
 
-![Piper screenshot](http://imgur.com/ELLCiSE.png)
+```
+$ flake8 --install-hook git
+$ git config --bool flake8.strict true
+```
```
bentiss: ... Anyway, flake8 will be run in the continuous integration while running your PR. So you'll get the result too while submitting the PR.

README.md
```diff
@@ -1,36 +1,78 @@
...
+
+Source
+======
+
+`git clone https://github.com/libratbag/piper.git`
```
bentiss: Naah

Do we really need that?

README.md
```diff
@@ -50,3 +92,12 @@ Licensed under the GPLv2
     You should have received a copy of the GNU General Public License
     along with this program; if not, write to the Free Software
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+See the [COPYING][copying] file for the full license information.
```
bentiss: Then you can remove most of the above ("Licensed under the GPLv2" should be enough)

README.md
```diff
@@ -50,3 +92,12 @@ Licensed under the GPLv2
...
+
+[ratbagd-readme]: https://github.com/libratbag/libratbag/blob/master/README.md#running-ratbagd-as-dbus-activated-systemd-service
+[meson]: http://mesonbuild.com/
+[ninja]: https://ninja-build.org/
+[kanban]: https://github.com/orgs/libratbag/projects/1
+[wiki-redesign]: https://github.com/libratbag/piper/wiki/Piper-Redesign
+[copying]: https://github.com/libratbag/piper/blob/master/COPYING
```
bentiss: Genuine question: I know we are not using doxygen for piper, but if we ever, would this links be handled properly?

Also, isn't that simpler to have them directly in the text because when I clone the repo, I read the text through vim^Wmy-favorite-text-editor* and it's easier to have the link just there.

\* not raising any trolls here, though vim is better than anything else

README.md
whot: yeah, it's easier to just link to the top where it has all the import statements...

technically, libratbag isn't even a dependency, piper is just a tad useless without it. But you do get to see a pretty mouse trap :)

README.md
whot: yeah, just cut it down to something that's still applicable in 3 months time, we don't look at the README often enough

README.md
whot: yeah, it's three lines and doesn't hurt. sorry, four lines. ok, 3 and a half, because it's whitespace only.

README.md
Hjdskes: True, but it goes faster for everyone involved if you run flake8 locally before opening a PR.

README.md
Hjdskes: @whot Are you saying we do or do not need it? I don't follow :P

README.md
Hjdskes: In that case the same applies for [libratbag](https://github.com/libratbag/libratbag#license) ;)

README.md
Hjdskes: Not sure if they'd be handled correctly, but I find the plain-text markdown easier to read this way. Will change it though

README.md
Hjdskes: Fixed

README.md
Hjdskes: Fixed

README.md
Hjdskes: Fixed
 Styling with Markdown is supported Comment Cancel
Show outdated

README.md
Hjdskes: Fixed

README.md
whot: we should have a link to the repo. main reason is also that with github forks, you don't necessarily know which one is the canonical upstream, having it listed in the README helps

PR #20, opened 2017-06-30
MouseMap: custom container widget to map the device SVG

Hjdskes:
This widget implements a central part of the new UI. See [this webm](https://hjdskes.github.io/img/blog/gsoc-part-4/mousemap-interactive.webm) to get an impression for that it does (note that the styling of the highlight is outdated; it now highlights the entire button instead of just its border).

Next up will be a new window making use of this widget, stealing code from piper.py.

### DIFF
--- a/piper/mousemap.py
+++ b/piper/mousemap.py
@@ -1,0 +1,387 @@
+# Copyright (C) 2017 Jente Hidskes <hjdskes@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import cairo
+import gi
+import os
+import sys
+from lxml import etree
+
+gi.require_version("Gdk", "3.0")
+gi.require_version("Gtk", "3.0")
+gi.require_version("Rsvg", "2.0")
+from gi.repository import Gdk, GLib, Gtk, GObject, Rsvg
+
+"""This module contains the MouseMap widget (and its helper class
+_MouseMapChild), which is central to the button and LED configuration stack
+pages. The MouseMap widget draws the device SVG in the center and lays out a
+bunch of child widgets relative to the leaders in the device SVG."""
+
+
+class _MouseMapChild:
+    # A helper class to manage children and their properties.
+
+    def __init__(self, widget, is_left, svg_id):
+        self._widget = widget
+        self._is_left = is_left
+        self._svg_id = svg_id
+        self._svg_leader = svg_id + "-leader"
+
+    @property
+    def widget(self):
+        # The widget belonging to this child.
+        return self._widget
+
+    @property
+    def svg_id(self):
+        # The identifier of the SVG element with which this child's widget is
+        # paired.
+        return self._svg_id
+
+    @property
+    def svg_leader(self):
+        # The identifier of the leader SVG element with which this child's
+        # widget is paired.
+        return self._svg_leader
+
+    @property
+    def is_left(self):
+        # True iff this child's widget is allocated to the left of the SVG.
+        return self._is_left
+
+
+class MouseMap(Gtk.Container):
+    """A Gtk.Container subclass to draw a device SVG with child widgets that
+    map to the SVG. The SVG should have objects with identifiers, whose value
+    should also be given to the `add` method. See `do_size_allocate` and
+    https://github.com/libratbag/libratbag/blob/master/data/README.md for more
+    information.
+    """
+
+    __gtype_name__ = "MouseMap"
+
+    __gproperties__ = {
+        "spacing": (int,
+                    "spacing",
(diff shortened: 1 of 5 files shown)


Hjdskes:
If you agree with this approach I can finalize the SVG guidelines in libratbag, and ask Jakub to do some more devices for us.

Hjdskes:
I just noticed that there is a 1 or 2 pixel offset with the right-aligned buttons. I think this is a rounding error but I'm not sure where it comes from.

For example, with a spacing and border width of 30, here is how this should align:

Left aligned buttons: x=30, width=95, sum=125
Spacing: x=125, width=30, sum=155
SVG: x=155, width=468, sum=623
Spacing: x=623, width=30, sum=653
Now, right aligned children should have x=653 but they have 651 instead.

This also happens for SVGs without left-aligned children, so I suspect it has something to do with librsvg. According to the documentation, it returns ints but if it does some rounding internally that could cause the offset. I'm going to look into this now but wanted to post this first so you won't merge it before I'm done.

Hjdskes:
To be explicit about types I rounded the returned floats to their nearest integer in `_translate_to_origin`, but I cannot fix the 1 or 2 pixel offset due to rounding. It's only visible with no spacing and no border width; the more of those the less visible it gets.

piper/mousemap.py
```diff
@@ -0,0 +1,390 @@
...
+        self._device = ratbagd_device
+        self._children = []
+        self._highlight_element = None
+
+        self._handle = Rsvg.Handle.new_from_file(ratbagd_device.svg_path)
+        self._svg_data = etree.parse(ratbagd_device.svg_path)
+
+    def do_add(self, widget):
+        """Not implemented, use `add(widget, svg_id)` instead."""
+        pass
```
whot: should we throw an exception here? seems like a programming error if that is ever called

piper/mousemap.py
```diff
@@ -0,0 +1,390 @@
...
+        width, the natural child widths (left and right), spacing and border
+        width.
+        """
+        width = 2 * self.props.border_width
+        width_svg = self._handle.props.width
+        width_left = max((child.widget.get_preferred_width()[1] for
+                          child in self._children if child.is_left), default=0)
+        width_right = max((child.widget.get_preferred_width()[1] for
+                          child in self._children if not child.is_left),
+                          default=0)
```
whot: nitpick: split the list comprehension up and make this two lines each, it'll be a lot easier to read if you have a
````
left_children = [foo for foo in foo]
width_left = max(left_children)
````

File: piper/mousemap.py [hunk 3]
```diff
@@ -0,0 +1,390 @@
...
+            'sodipodi': 'http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd',
+            'cc': 'http://web.resource.org/cc/',
+            'svg': 'http://www.w3.org/2000/svg',
+            'dc': 'http://purl.org/dc/elements/1.1/',
+            'xlink': 'http://www.w3.org/1999/xlink',
+            'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+            'inkscape': 'http://www.inkscape.org/namespaces/inkscape'
+        }
+        query = "//svg:rect[@id=\"{}\"][contains(@style, \"{}\")]".\
+                format(svg_id, style)
```
whot: screw the 78 char limit here and make this one line please, much more obvious that way

piper/mousemap.py
```diff
@@ -0,0 +1,390 @@
...
+    def _translate_to_origin(self):
+        # Translates the coordinate system such that the SVG and its buttons
+        # will be drawn in the center of the allocated space. The returned x-
+        # and y-coordinates will be the top left corner of the centered SVG.
+        allocation = self.get_allocation()
+        width = self.get_preferred_width()[1]
+        height = self.get_preferred_height()[1]
+
+        width_left = max((child.widget.get_preferred_width()[1] for
+                          child in self._children if child.is_left), default=0)
```
whot: same here, please split it over two lines

whot:
Just a few nitpicks, please push when they're addressed

piper/mousemap.py
Hjdskes: Perhaps it is better to simply not implement this virtual/interface method at all? In that case, it won't be listed in the documentation (can also do this by changing the docstring to a comment), and if someone tries to call this method on our container anyway they'll fall through to https://github.com/GNOME/gtk/blob/master/gtk/gtkcontainer.c#L1575

Hjdskes: In fact, Python will automatically reference the `add(widget, svg_id)` method: 

```
 > ./piper.in
  File "/home/jente/code/src/github.com/libratbag/piper/piper/application.py", line 53,
in do_activate
    window = Window(self, self._ratbag)
  File "/home/jente/code/src/github.com/libratbag/piper/piper/window.py", line 102, in _
_init__
    stack.add_titled(self._setup_buttons_page(device), "buttons", _("Buttons"))
  File "/home/jente/code/src/github.com/libratbag/piper/piper/window.py", line 145, in _
setup_buttons_page
    mousemap.add(mapbutton)
TypeError: add() missing 1 required positional argument: 'svg_id'
```

To use the virtual `do_add` method and fall through to the line in my earlier comment, the user (developer) will have to explicitly call `mousemap.do_add(mousemap, widget)` which is unlikely to happen.

So I'll just remove `do_add` and we should be fine.

Hjdskes:
@whot (and others): I just pushed a skeleton window that makes it easier to test the changes. Should you want to test them, make sure you download [these SVGs](https://gist.github.com/Hjdskes/b96b351ae0920d56959de084d45d53fa) that have been modified according to the [new guidelines](https://github.com/Hjdskes/libratbag/blob/3d44c6930592cab42c22862c713a10be22b894cb/data/README.md). To make it easy to test with these SVGs, modify lines [114 and 115](https://github.com/Hjdskes/piper/blob/c216782895d192266849e42cf4707a013a4745d2/piper/mousemap.py#L114..L115) to use absolute paths to your downloaded SVGs.

PR #26, opened 2017-07-05
Add error codes to the ratbagd bindings, plus commit method

Hjdskes:
A quick PR that adds two things: a `Commit` method on `RatbagdDevice`, and libratbag's error codes so that Piper can easily test return values of the DBus method calls.

I've though about adding exceptions as well (i.e. `RatbagdErrorDevice(BaseException)`) but this gets nasty real quick; either every method call in the binding needs to do an `if/else` to raise the proper exception, or `_dbus_call` does this but then has the possibility of raising *all* exceptions which isn't nice from an API point of view.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -81,13 +81,16 @@ class _RatbagdDBus(GObject.GObject):
         if self._proxy.get_name_owner() is None:
             raise RatbagdDBusUnavailable()
 
-    def dbus_property(self, property):
+    def _dbus_property(self, property):
+        # Retrieves a cached property from the bus, or None.
         p = self._proxy.get_cached_property(property)
         if p is not None:
             return p.unpack()
         return p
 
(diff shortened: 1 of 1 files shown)


piper/ratbagd.py
```diff
@@ -81,18 +81,21 @@ def __init__(self, interface, object_path):
...
-    def dbus_call(self, method, type, *value):
+    def _dbus_call(self, method, type, *value):
+        # Calls a method synchronously on the bus, using the given method name,
+        # type signature and values. Returns the returned result, or None.
         val = GLib.Variant("({})".format(type), value)
         res = self._proxy.call_sync(method, val,
                                     Gio.DBusCallFlags.NO_AUTO_START, 500, None)
         if res is not None:
-            return res.unpack()
+            return res.unpack()[0]  # Result is always a tuple
```
whot: this needs to be in a separate patch, it mustn't be hidden in a rename patch

PR #28, opened 2017-07-06
Add resolutions page

Hjdskes:
This PR adds the resolutions page according to the mockups (well, mostly). It's not entirely done but I would like some feedback and discussion:

1. Libratbag doesn't yet support adding and removing resolutions (see https://github.com/libratbag/libratbag/issues/211), how do I deal with this in Piper? At the moment, the widgets are there printing a message to stdout.
2. To draw only one leader in the MouseMap in the resolutions page, I modified the SVG requirements (again...). I'm not particularly happy with this as the list of SVG requirements is already pretty long, but basically I grouped the paths and the little squares they connect to and gave this group the id `buttonX-path`. Commit https://github.com/libratbag/piper/pull/28/commits/fb068e395c076755a2c2535c12e5fa842843c0d1 should explain itself, I think. If we don't do this, the MouseMap in the resolutions page will draw all the leaders, even those without a "Switch resolution" label next to it.
3. There is no option to set a resolution as the default; how shall I approach this? The mockups left this out.

I think that's all. Do try it out (any SVG from https://github.com/libratbag/libratbag/pull/182 should work), and let me know what you think of it thus far. It would also be great if someone can test it with a device that supports separate X/Y resolutions, as I cannot do this myself.

### DIFF
--- a/data/piper.gresource.xml
+++ b/data/piper.gresource.xml
@@ -5,6 +5,7 @@
 
         <file preprocess="xml-stripblanks">aboutDialog.ui</file>
         <file preprocess="xml-stripblanks">resolutionRow.ui</file>
+        <file preprocess="xml-stripblanks">resolutionsPage.ui</file>
         <file preprocess="xml-stripblanks">window.ui</file>
 	<!-- Using this alias, GtkApplication will automatically pick it up for us. -->
         <file alias="gtk/menus.ui" preprocess="xml-stripblanks">menus.ui</file>
--- a/data/piper.gresource.xml
+++ b/data/piper.gresource.xml
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <gresources>
-	<gresource prefix="/org/freedesktop/Piper">
-		<file preprocess="xml-stripblanks">aboutDialog.ui</file>
-		<file>404.svg</file>
-		<!-- Using this alias, GtkApplication will automatically pick it up for us. -->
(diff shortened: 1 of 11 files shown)


whot:
General comments to your questions, I'll skip 1) because it's in the libratbag issue.

2) I think grouping them is a good idea and fairly easy to do in inkscape. I think this is just a case of where an SVG evaluation tool needs to catch this early, that way it's the least annoying.

3) Might be a question for @jimmac, but how about a little tick next to the (x) that works like a radiobutton? it's not really intuitive. The other option would be to have an 'advanced' button that pops up something like the breathing dialog in https://github.com/libratbag/piper/raw/wiki/redesign/leds.png. That dialog could handle the uneven x/y resolutions and include a checkbox for "make this the default resolution". This way you could also switch to equal x/y resolutions in the main window

Hjdskes:
@whot I fixed the issues you mentioned on IRC (scrolling over scales stops the scrolling in the list, separate X/Y resolutions being a niche case, hiding configuration when opening another and the scale snapping to multiples of 50). I also extracted the code for the resolutions stack page into its own class.

As for point 2, I updated the instructions in https://github.com/libratbag/libratbag/pull/182 and the current three SVGs in there. That leaves only point 3 (assuming we'll do point 1 later when this is done in libratbag).

whot:
We can track the design for 3 in a separate issue, I think this branch is good enough to merge. Let's do it before it gets too unwieldy or before we get side-tracked.

PR #34, opened 2017-07-11
data: move ui files into subdirs

Hjdskes:
It's getting quite messy to let's put in some structure.


whot:
happy moving it to ui but is the extra nesting necessary? seems a bit excessive for what will be less than 20 files in the end

Hjdskes:
I suppose that's personal preference, I'm happy to go either way.

whot:
i think it'll be easier if we have them in the ui/ folder without any further nesting unless the number of them gets out of hand. Feel free to push directly once you've flattened that bit

PR #35, opened 2017-07-11
Add LED page to configure LEDs

Hjdskes:
As per [the mockups](https://github.com/libratbag/piper/raw/wiki/redesign/leds.png), but dialogs instead of popovers to consistently use dialogs throughout Piper (as they are required for button mapping) and with the color selection button moved into the configuration dialog as not every mode has a configurable color.

Changes are applied only after the commit button is pressed. State in between two invocations of any configuration dialog is not remembered; not even after a commit. The former is intentional, the latter is a bug. I suspect this is due to the bindings only retrieving their properties from DBus upon instantiation, but I'll look into this after opening this PR.

Also, as noted in the code, we cannot toggle *after* the user has accepted the changes in the dialog or not toggle when the user cancelled the dialog. This means that the mode will be changed even when the user cancels that mode's configuration dialog. I don't like it, it would certainly be better to not toggle when the user cancels the dialog, but I suppose it's fine since the changes have to be committed before they are applied to the device. If it's a deal breaker, I can look into subclassing `Gtk.ToggleButton` to fix this. Let me know your opinion.

Finally, I attempted to add capability checking before adding pages to the main window's stack, but I'm not sure if I'm doing this correctly; my device doesn't seem to have `RatbagdDevice.CAP_RESOLUTION` (which I think it should, given its documentation) but it does have `CAP_SWITCHABLE_RESOLUTION`. Which of these should I use to check if the resolutions page should be shown?

I think that's it; please test the changes and let me know what you think.

### DIFF
--- a/piper/leddialog.py
+++ b/piper/leddialog.py
@@ -1,0 +1,88 @@
+# Copyright (C) 2017 Jente Hidskes <hjdskes@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+from .gi_composites import GtkTemplate
+from .ratbagd import RatbagdLed
+
+import gi
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gdk, GObject, Gtk
+
+
+@GtkTemplate(ui="/org/freedesktop/Piper/ui/LedDialog.ui")
+class LedDialog(Gtk.Dialog):
+    """A Gtk.Dialog subclass to implement the dialog that shows the
+    configuration options for the LED effects."""
+
+    __gtype_name__ = "LedDialog"
+
+    stack = GtkTemplate.Child()
+    colorchooser = GtkTemplate.Child()
+    colorbutton = GtkTemplate.Child()
+    adjustment_brightness = GtkTemplate.Child()
+    adjustment_effect_rate = GtkTemplate.Child()
+
+    def __init__(self, ratbagd_led, *args, **kwargs):
+        """Instantiates a new LedDialog.
+
+        @param ratbagd_led The LED to configure, as ratbagd.RatbagdLed.
+        """
+        Gtk.Dialog.__init__(self, *args, **kwargs)
+        self.init_template()
+        self._led = ratbagd_led
+        self._modes = {
+            "solid": RatbagdLed.MODE_ON,
+            "cycle": RatbagdLed.MODE_CYCLE,
+            "breathing": RatbagdLed.MODE_BREATHING,
+            "off": RatbagdLed.MODE_OFF
+        }
+
+        mode = self._led.mode
+        for k, v in self._modes.items():
+            if mode == v:
+                self.stack.set_visible_child_name(k)
+        rgba = self._get_led_color_as_rgba()
+        self.colorchooser.set_rgba(rgba)
+        self.colorbutton.set_rgba(rgba)
+        self.adjustment_brightness.set_value(self._led.brightness)
+        self.adjustment_effect_rate.set_value(self._led.effect_rate)
+
+    def _get_led_color_as_rgba(self):
+        # Helper function to convert ratbagd's 0-255 color range to a Gdk.RGBA
+        # with a 0.0-1.0 color range.
+        r, g, b = self._led.color
+        return Gdk.RGBA(r / 255.0, g / 255.0, b / 255.0, 1.0)
+
+    @GObject.Property
+    def mode(self):
+        visible_child = self.stack.get_visible_child_name()
+        return self._modes[visible_child]
+
+    @GObject.Property
+    def color(self):
+        if self.mode == RatbagdLed.MODE_ON:
+            rgba = self.colorchooser.get_rgba()
+        else:
+            rgba = self.colorbutton.get_rgba()
+        return (rgba.red * 255.0, rgba.green * 255.0, rgba.blue * 255.0)
+
+    @GObject.Property
+    def brightness(self):
+        return self.adjustment_brightness.get_value()
+
+    @GObject.Property
+    def effect_rate(self):
+        return self.adjustment_effect_rate.get_value()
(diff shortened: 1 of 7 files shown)


Hjdskes:
> I suspect this is due to the bindings only retrieving their properties from DBus upon instantiation, but I'll look into this after opening this PR.

Always retrieving a property over DBus doesn't seem to be enough; it seems we also need to explicitly call `proxy.set_cached_property(property_name, value)` when setting properties. Am I correct in assuming this?

piper/ledbutton.py
```diff
@@ -0,0 +1,181 @@
...
+        self.button_breathing.connect("clicked", self._on_button_clicked, RatbagdLed.LED_MODE_BREATHING)
+
+    def _on_button_clicked(self, button, mode):
+        # Responds to a click on the button. The button internally always
+        # toggles on a click, so we cannot toggle only when the user applies
+        # changes in the dialog and abort when the user cancels the dialog. As
+        # such, we always change the mode; this is probably fine since the user
+        # has to commit the changes to the device anyway, but it's not ideal.
+        if not button.get_active():
+            # Prevent the unactivated togglebutton's dialog from showing.
```
whot: that confused me, you mean the unsupported ones, right?

piper/ledbutton.py
```diff
@@ -0,0 +1,181 @@
...
+            return
+
+        self._led.mode = mode
+        if mode == RatbagdLed.LED_MODE_ON:
+            self._open_solid_dialog()
+        elif mode == RatbagdLed.LED_MODE_CYCLE or mode == RatbagdLed.LED_MODE_BREATHING:
+            self._open_advanced_dialog(mode)
```
whot: cache the currently active one and get the result of the dialog check for cancel. If cancel, you could re-toggle the currently active one? Not sure whether that'll flicker though.

whot:
ok, looking at the mockups again and my device here I'm thinking the best option may be to go the same route as the button page. Have a little gear icon that pops up a full dialog. In that dialog you can have the 4 types as toggle buttons like you have them now, with the settings for each of them appearing below as you toggle. I think that's the most conservative approach and will scale nicely for mice with lots of LEDs.

piper/ledbutton.py
Hjdskes: I mean that if button A is toggled, and now you toggle button B, then both A's and B's dialogs appear. This is because both A and B emit the `toggled` signal; A because it was deactivated and B because it was activated. This check is there so that a dialog only appear if a button is activated from the toggle, i.e. so that only button B's dialog appears.

Hjdskes: So I guess s/unactivated/deactivated/

whot: maybe use "ignore buttons toggled off" or something?

Hjdskes: Not relevant anymore in the new design :)

Hjdskes:
Oke, attempt nr 2. I went with @whot's suggestion to go the same route as the button page.

State is now remembered in between dialog invocations. We still always set the mode, even if the dialog is cancelled. I'll look into this later. @whot Can I get some feedback on the last point raised in the opening comment?

Let me know what you think!

Hjdskes:
Always setting the mode (and other settings) is a trade off between having the same state in between dialog invocations and, well, only applying the changes when the user presses the Apply button (note that applying != committing, that remains a separate button press). Before I get to work on this, what are your preferences?

whot:
Ok, the capability bits are fixed with the PR above, at least for some of the drivers. The meaning between the two differs, RESOLUTION just means that you can read and/or write resolutions, SWITCHABLE means that you can switch between them somehow. In theory, there could be a mouse that only has one resolution but not SWITCHABLE, so for you the RESOLUTION capability matters.

Having said that, we should introduce a RESOLUTION_WRITABLE flag to indicate we can change it in software. There are a few devices that have multiple resolutions but only toggleable via a hw switch and they cannot be changed. They don't really matter much to libratbag though.

whot:
> Always setting the mode (and other settings) is a trade off between having the same state in between dialog invocations and, well, only applying the changes when the user presses the Apply button (note that applying != committing, that remains a separate button press). Before I get to work on this, what are your preferences? Personally I like what's there currently.

sorry, I'd rather have the mode *not* change on cancel. I played with it a bit, and it's the classic case of letting the user panic without side-effects. This scenario: 
- "ooh, I wonder what the LEDs do" (clicks button, dialog comes up). 
- "What is breathing?" (clicks button 'breathing')
- "Whoah, I don't know what this does, let's get out of here" (clicks cancel). 
In that case, the user still switched to breathing mode, which is different to how dialogues behave in the rest of the system. So I think we really should ignore any changes after cancel.

piper/leddialog.py
```diff
@@ -0,0 +1,106 @@
...
+
+        mode = self._led.mode
+        if mode == RatbagdLed.MODE_ON:
+            self.stack.set_visible_child_name("solid")
+        elif mode == RatbagdLed.MODE_CYCLE:
+            self.stack.set_visible_child_name("cycle")
+        elif mode == RatbagdLed.MODE_BREATHING:
+            self.stack.set_visible_child_name("breathing")
+        elif mode == RatbagdLed.MODE_OFF:
+            self.stack.set_visible_child_name("off")
```
whot: I think you could/should use a dictionary [string] = enum here for the mode mappings. In `_on_stack_page_changed` you can do a simple lookup, here you either invert the dict or simply loop through `.keys()` until you find the value.

piper/leddialog.py
```diff
@@ -0,0 +1,106 @@
...
+        elif visible_child == "breathing":
+            self._led.mode = RatbagdLed.MODE_BREATHING
+        elif visible_child == "off":
+            self._led.mode = RatbagdLed.MODE_OFF
+
+    def _get_led_color_as_rgba(self):
+        # Helper function to convert ratbagd's 0-255 color range to a Gdk.RGBA
+        # with a 0.0-1.0 color range.
+        r, g, b = self._led.color
+        return Gdk.RGBA(r / 255, g / 255, b / 255, 1.0)
```
whot: This is me having been bitten by C and Python2 too many times: use `255.0` here please to make it obvious that it's not an integer division. I know it's not in Py3, but still...

Also, we don't usually do spaces around /, afair

Hjdskes: flake8 complains about `/` without spaces, so I guess we'll stick with that.

Hjdskes:
With the latest additions, every issue raised in this PR should be fixed. Please test it again.

I added the two commits from #41 in here, so we should finish that before merging this.

piper/ratbagd.py
```diff
@@ -501,6 +501,7 @@ def mode(self, mode):
         @param mode The new mode, as one of MODE_OFF, MODE_ON, MODE_CYCLE and
                     MODE_BREATHING.
         """
+        print("Setting mode to:", mode)
```
whot: this looks like a leftover

Hjdskes: Oversight of the cherry pick indeed.

Hjdskes:
This is ready to merge now, as soon as #41 is in.

Issue #31, opened 2017-07-10
Resolutions page needs a way to set a profile's default resolution.


Hjdskes:
The mockups left this out, so this isn't yet decided.

@whot [suggested](https://github.com/libratbag/piper/pull/28#issuecomment-313603601) a check mark next to the delete button (functioning like a radio button, e.g. only one can be selected at any given time) or an "advanced" popup menu that has the check mark in there, and if supported by the device also exposes the scale for the y-resolution. My personal preference would be the former, but let's open a discussion on this.

bentiss:
I also would prefer to minimize the interactions, and thus having a radio button next to the resolutions seem IMO the best option

whot:
cc @jimmac

Hjdskes:
In progress in #36; please move the discussion there.

PR #36, opened 2017-07-11
ResolutionPage: add default resolution button

Hjdskes:
As suggested in #31 a radio button to set the default resolution.

One thing I noticed is that the DBus signals (at least for the resolutions, e.g. `ActiveResolutionChanged` -- this one is [expected](https://github.com/libratbag/libratbag/pull/208) not to fire -- and `DefaultResolutionChanged`) don't seem to fire, but I haven't investigated this as currently Piper assumes it's the only one changing values on the bus...

### DIFF
--- a/data/ui/ResolutionRow.ui
+++ b/data/ui/ResolutionRow.ui
@@ -14,41 +14,9 @@
             <property name="can_focus">False</property>
             <property name="border_width">12</property>
             <child>
-              <object class="GtkLabel" id="index_label">
-                <property name="visible">True</property>
-                <property name="can_focus">False</property>
-                <property name="valign">baseline</property>
-                <property name="single_line_mode">True</property>
-                <property name="track_visited_links">False</property>
-                <attributes>
-                  <attribute name="weight" value="bold"/>
-                </attributes>
-              </object>
-              <packing>
-                <property name="expand">False</property>
-                <property name="fill">True</property>
-                <property name="position">0</property>
-              </packing>
-            </child>
-            <child>
-              <object class="GtkLabel" id="title_label">
-                <property name="visible">True</property>
-                <property name="can_focus">False</property>
-                <property name="halign">center</property>
-                <property name="justify">center</property>
-                <property name="track_visited_links">False</property>
-              </object>
-              <packing>
-                <property name="expand">True</property>
-                <property name="fill">True</property>
-                <property name="position">1</property>
-              </packing>
-            </child>
-            <child>
               <object class="GtkButton" id="delete_button">
                 <property name="can_focus">False</property>
                 <property name="receives_default">False</property>
-                <property name="tooltip_text" translatable="yes">Remove this resolution from the profile</property>
                 <property name="relief">none</property>
                 <signal name="clicked" handler="_on_delete_button_clicked" swapped="no"/>
                 <child>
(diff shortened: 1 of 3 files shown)


Hjdskes:
> sorry, I'm gonna say 'no' on this

No worries :) I'll try your suggestions after fixing the LEDs page, which has a higher priority for me right now.

Hjdskes:
Right, I now remember why I went with radio buttons. No other widget enforces that at all times there is exactly one option selected; i.e. say we're using a [GtkSwitch](https://developer.gnome.org/gtk3/stable/GtkSwitch.html), then what is supposed to happen when the user unsets the default resolution? That'd mean there is *no* default resolution, which I assume isn't something we want to expose.

Also, anything other than radiobuttons will require the emission of signals to work, so until we get that fixed I cannot test anything. The ratbagd bindings seem to be correct; I see nothing in `dbus-monitor` when I change the default resolution but it *is* getting set.

whot:
To rule out the obvious: you're using dbus-monitor --system? It defaults to session, caught me out at first.

whot:
> say we're using a GtkSwitch, then what is supposed to happen when the user unsets the default resolution? That'd mean there is no default resolution, which I assume isn't something we want to expose.

Two options: either make the switch insensitive when active so you can only ever enable it, but not disable it. But I think a checkbox may be better here anyway, it seems better for a "tick this one" option.

The other option would be that as you untick the currently active one, you randomly pick another one that becomes default. But I think the first option is better, even though it's not great UX either.

Hjdskes:
> To rule out the obvious: you're using `dbus-monitor --system`? It defaults to session, caught me out at first.

Thanks, that was it. The signal is indeed being fired, so the issue must be in the bindings somewhere.

> Two options: either make the switch insensitive when active so you can only ever enable it, but not disable it. But I think a checkbox may be better here anyway, it seems better for a "tick this one" option.
>
> The other option would be that as you untick the currently active one, you randomly pick another one that becomes default. But I think the first option is better, even though it's not great UX either.

I also prefer the first option, including a checkbox instead of the switch we have now. I'll have a go at it later today.

Hjdskes:
I think the signal isn't being received by the bindings because a `RatbagdResolution` exists for each resolution, i.e. has object path `/org/freedesktop/ratbag1/resolution/event13/p0/r0` with interface `org.freedesktop.ratbagd1.Resolution`, while the `DefaultResolutionChanged` signal is emitted on `/org/freedesktop/ratbag1` and with interface `org.freedesktop.ratbagd1.Resolution`. GDBusProxy says the following for its `g-signal` signal: "Emitted when a signal from the remote object and interface that proxy is for, has been received." So because the signal it receives isn't for its object path (`.../resolution/event13/p0/r0`) it doesn't emit its `g-signal`.

I'm not entirely sure how I can fix this (not that familiar with sd-bus) but I'll have a go.

Hjdskes:
That was easier than expected, hehe. See https://github.com/libratbag/libratbag/pull/226.

Hjdskes:
Again CircleCI fails with a dependency error:

```
BDB1539 Build signature doesn't match environment
failed loading RPMDB
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Exited with code 1
```

whot:
> Again CircleCI fails with a dependency error:

caused by rawhide, I have now switched to F26 instead, see https://github.com/libratbag/libratbag/pull/227

Hjdskes:
The last commits show a new attempt after the DBus interface changes of this week. It's almost working now, except for the fact that the old default resolution's `IsDefault` property isn't being detected by `RatbagdResolution`'s DBus proxy. As a result, there are always two default resolutions (the prints are still in the code so you can follow and/or try it yourself):

```
 > ./build/piper.devel

Toggled on resolution 0
notify::is-default on resolution 0 True
  New default
notify::is-default on resolution 1 True
  New default
notify::is-default on resolution 2 False
notify::is-default on resolution 3 False
notify::is-default on resolution 4 False
```

The new default works because we can set its `IsDefault` property on the proxy immediately, but we cannot do this for the old one and so we have the same race condition as in 7fd57b2bd3c3c61dc9252b69fd7db1f7859f252d.

Commit e9ea7ddca77807af191464029367deb9e8a96dda should really be done on master instead but I'll leave it here for now until this is sorted out.

whot:
I think that's something where we have to add special knowledge to the ratbag bindings so sync all when syncing that property. Not sure there's a good way around it without triggering the race conditions.

Hjdskes:
> I think that's something where we have to add special knowledge to the ratbag bindings so sync all when syncing that property.

How would this look? AFAIK one `RatbagResolution` cannot access the others.

whot:
No, but we can add that code :)

whot:
closing this for now because it's still in WIP state but hasn't seen any active updates in a few days. Let's re-open it when we're ready

PR #37, opened 2017-07-12
Always query DBus for the latest value in the bindings

Hjdskes:
See #35 for why, but basically this is just the Right Thing™

Sidenote: I should really start writing some tests for these, especially with the introduction of the driver-test libratbag element from https://github.com/libratbag/libratbag/pull/218

### DIFF
--- a/piper/piper.py
+++ b/piper/piper.py
@@ -247,7 +247,7 @@ class Piper(Gtk.ApplicationWindow):
         for l, button in zip(self._button_function_labels, buttons):
             action = button.action_type
             if action == "button":
-                text = "Button {} click".format(button.button_mapping)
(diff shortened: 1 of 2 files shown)


Hjdskes:
> Given that this allows to reduce the testsuite for ratbagctl from 22 to 2 secs, I am approving and merging this :)

Where does this improvement come from?

bentiss:
> Where does this improvement come from?

It's just that now I do not need to spawn a new process for each call, which creates a parser and does all the dbus initialization.
Before this, the ratbagd binding kept its internal state meaning a change and a query in the same process means that we would not update the internal state correctly.

bentiss:
Though I must say this new binding allows to divide the time of the test suite by 2 (from 20 seconds to 10)

PR #39, opened 2017-07-12
ratbagd.py: sync with libratbag

bentiss:
This is a convenient way for me to sync between the 2, but it will
allow piper to have tests too :)

Note I haven't properly tested this in piper itself, counting on you, @Hjdskes :)


Hjdskes:
Everything seems to be working fine still, thanks!

PR #41, opened 2017-07-13
Explicitly and immediately set cached DBus properties on update

Hjdskes:
This is required to explicitly and immediately update the cached properties, to prevent race conditions from occurring in Piper.

I didn't connect to the [`g-properties-changed`](https://developer.gnome.org/gio/stable/GDBusProxy.html#GDBusProxy-g-properties-changed) signal, as according to the linked documentation the "local cache has already been updated when this signal fires", AKA: this won't solve the race condition, at least not if I understand the documentation correctly.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -145,6 +145,11 @@ class _RatbagdDBus(GObject.GObject):
             return p.unpack()
         return p
 
+    def _set_dbus_property(self, property, type, value):
+        # Sets a cached property on the bus.
+        val = GLib.Variant("{}".format(type), value)
+        self._proxy.set_cached_property(property, val)
+
     def _dbus_call(self, method, type, *value):
         # Calls a method synchronously on the bus, using the given method name,
         # type signature and values.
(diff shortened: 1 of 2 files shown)


Hjdskes:
CircleCI fails on the install step, seems like a dependency isn't installing correctly.

It just occurred to me that we may only want to update the cached property when the DBus call (e.g. `SetMode` for an LEDs mode) succeeds.

whot:
ok, sorry, I didn't see your last comment before I reviewed it...

Hjdskes:
> shouldn't we only update the cached property if the dbus call didn't fail?

Done

piper/ratbagd.py
```diff
@@ -59,6 +59,46 @@ class RatbagdDBusUnavailable(BaseException):
...
+
+
+class RatbagErrorSystem(BaseException):
+    """An exception corresponding to RatbagErrorCode.RATBAG_ERROR_System."""
+    pass
+
+
+class RatbagErrorImplementation(BaseException):
+    """An exception corresponding to RatbagErrorCode.RATBAG_ERROR_IMPLEMENTATION."""
+    pass
```
whot: might want to make a general RatbagError and then derive from that one, just so callers can have a `except RatbagError` clause to catch all the specific ones. But we can add that later.

PR #43, opened 2017-07-13
Rename anonymous variables from _ to __

Hjdskes:
Modules that need internationalization import gettext as `_`, so anonymous variables should be named `__` instead (double underscore) so as not to clash.


Hjdskes:
This change is so small that I'll just merge it directly. Feel free to open an issue or comment here if you disagree.

PR #46, opened 2017-07-14
data/ui: add tooltips to control widgets

### LINKED ISSUE #40: Control widgets should have tooltips to explain their purpose
As per the title.

Hjdskes:
Every control widget should have a tooltip to explain its purpose.

Fixes #40

### DIFF
--- a/data/ui/Window.ui
+++ b/data/ui/Window.ui
@@ -8,6 +8,20 @@
       <object class="GtkOverlay">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
+        <child>
+          <object class="GtkStack" id="stack">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="transition_duration">400</property>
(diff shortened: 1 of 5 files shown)


data/ui/Window.ui
```diff
@@ -8,6 +8,20 @@
...
+            <property name="transition_duration">400</property>
+            <property name="transition_type">slide-left-right</property>
+            <child>
+              <placeholder/>
+            </child>
+          </object>
+          <packing>
+            <property name="index">-1</property>
+          </packing>
+        </child>
```
whot: This looks like an accidental change, can you undo  it before pushing please?

Hjdskes: This is Glade moving things around upon save; if I revert it now it will only do it again next time I edit this file. I think we should let it through tbh.

whot: just make it a separate commit then please, feel free to push directly

PR #47, opened 2017-07-17
Add ButtonPage to configure buttons

Hjdskes:
This PR is meant to track the implementation of the button stack page. The layout is according to the [mockup](https://github.com/libratbag/piper/raw/wiki/redesign/buttons.png), with the dialog similar to the current LED dialog (i.e., a stack page for each type of mapping, exposing the configuration for that particular type).

Each page will be implemented in its own separate commit for easy review; while this PR is WIP, each separate commit is considered ready for review.


piper/buttonpage.py
```diff
@@ -0,0 +1,72 @@
...
+            button = OptionButton(_("Button {}").format(index))
+            button.connect("clicked", self._on_button_clicked, ratbagd_button)
+            mousemap.add(button, "#button{}".format(index))
+            sizegroup.add_widget(button)
+
+    def _on_button_clicked(self, button, ratbagd_button):
+        # Presents the ButtonDialog to configure the mouse button corresponding
+        # to the clicked button.
+        profile = self._device.active_profile
+        n_buttons = len(profile.buttons)
```
Hjdskes: This part in particular I am not happy with, but it is the simplest approach for now. I could pass the complete profile instead of only the number of buttons, but I'm not sure how that will change once I add support for profiles. Although, similarly, I am not sure how this will change either -- I only know that this *will* change once I start on profiles.

piper/buttonspage.py
```diff
@@ -0,0 +1,73 @@
...
+            button = OptionButton(_("Button {}").format(index))
+            button.connect("clicked", self._on_button_clicked, ratbagd_button)
+            mousemap.add(button, "#button{}".format(index))
+            sizegroup.add_widget(button)
+
+    def _on_button_clicked(self, button, ratbagd_button):
+        # Presents the ButtonDialog to configure the mouse button corresponding
+        # to the clicked button.
+        profile = self._device.active_profile
+        n_buttons = len(profile.buttons)
```
Hjdskes: My previous comment disappeared due to a rebase, so I'll say it again:

I'm not happy with this approach; the dialog needs to know how many buttons the device has to populate the possible options for its button mapping page. Currently this is the simplest solution that *will* change once I implement profiles (i.e. when I figure out how to chain profiles through Piper in a nice way, that also allows for profile switching), but yea, suggestions welcome.

Hjdskes:
So currently this sports the button remapping portion, but it's not really intuitive; the dialog just presents a list of numbers that are assigned to buttons:

screenshot from 2017-07-17 15-11-30

Is there some way to put more descriptive actions here, such as "left click"? I suppose I could pass, instead of the number of buttons, all buttons in a list and retrieve those buttons' types instead of their numbers (see below). Then the list would be filled with e.g. "left", "middle", "right", "thumb", et cetera... but the problem is, that at least for my device, `type` always returns `unknown`. If it happens for my device, I don't think we can guarantee (even with libratbag under our control) that it won't happen for others either.

```diff
diff --git a/piper/buttondialog.py b/piper/buttondialog.py
index 5effbcb..2f0932a 100644
--- a/piper/buttondialog.py
+++ b/piper/buttondialog.py
@@ -31,21 +31,21 @@ class ButtonDialog(Gtk.Dialog):
     stack = GtkTemplate.Child()
     mapping_box = GtkTemplate.Child()
 
-    def __init__(self, ratbagd_button, n_buttons, *args, **kwargs):
+    def __init__(self, ratbagd_button, buttons, *args, **kwargs):
         """Instantiates a new ButtonDialog.
 
         @param ratbagd_button The button to configure, as ratbagd.RatbagdButton.
-        @param n_buttons The number of buttons on this device, as int.
+        @param buttons The buttons on this device, as [ratbagd.RatbagdButton].
         """
         Gtk.Dialog.__init__(self, *args, **kwargs)
         self.init_template()
         self._button = ratbagd_button
         self._mapping = ratbagd_button.mapping
 
-        self._init_ui(n_buttons)
+        self._init_ui(buttons)
 
-    def _init_ui(self, n_buttons):
-        self._init_mapping_page(n_buttons)
+    def _init_ui(self, buttons):
+        self._init_mapping_page(buttons)
 
         active_action_type = self._button.action_type
         action_types = self._button.action_types
@@ -55,29 +55,29 @@ class ButtonDialog(Gtk.Dialog):
             if active_action_type == action_type:
                 self.stack.set_visible_child_name(action_type)
 
-    def _init_mapping_page(self, n_buttons):
+    def _init_mapping_page(self, buttons):
         # Initializes the mapping stack page.
         # In libratbag, button mappings start at index 1 hence the offset in the
         # range. We use Gtk.RadioButton to ensure only one mapping is assigned
         # at any given time. Buttons whose type is not currently `mapping` won't
         # have any option active and are thus supported "for free".
         prev = None
-        for i in range(1, n_buttons + 1):
-            button = Gtk.RadioButton(i)
-            button.join_group(prev)
-            button.props.draw_indicator = False
-            button.set_visible(True)
-            if i == self._mapping:
-                button.set_active(True)
-            button.connect("toggled", self._on_mapping_button_toggled, i)
-            self.mapping_box.pack_start(button, True, True, 0)
-            prev = button
+        for button in buttons:
+            radio = Gtk.RadioButton(button.type)
+            radio.join_group(prev)
+            radio.props.draw_indicator = False
+            radio.set_visible(True)
+            if button.mapping == self._mapping:
+                radio.set_active(True)
+            radio.connect("toggled", self._on_mapping_button_toggled, button)
+            self.mapping_box.pack_start(radio, True, True, 0)
+            prev = radio
 
-    def _on_mapping_button_toggled(self, button, mapping):
+    def _on_mapping_button_toggled(self, radio, button):
         # A mapping radiobutton has been toggled, so we update our cached
         # mapping value. Note that by virtue of using a radiobutton, the new
         # mapping is always guaranteed to be different from the old.
-        self._mapping = mapping
+        self._mapping = button.mapping
 
     @GObject.Property
     def action_type(self):
diff --git a/piper/buttonspage.py b/piper/buttonspage.py
index ede0b66..fc88f47 100644
--- a/piper/buttonspage.py
+++ b/piper/buttonspage.py
@@ -58,9 +58,8 @@ class ButtonsPage(Gtk.Box):
     def _on_button_clicked(self, button, ratbagd_button):
         # Presents the ButtonDialog to configure the mouse button corresponding
         # to the clicked button.
-        profile = self._device.active_profile
-        n_buttons = len(profile.buttons)
-        dialog = ButtonDialog(ratbagd_button, n_buttons, transient_for=self.get_toplevel())
+        buttons = self._device.active_profile.buttons
+        dialog = ButtonDialog(ratbagd_button, buttons, transient_for=self.get_toplevel())
         dialog.connect("response", self._on_dialog_response, ratbagd_button)
         dialog.present()
```

piper/buttonspage.py
Hjdskes: See also https://github.com/libratbag/piper/pull/47#issuecomment-315753388

whot:
> but the problem is, that at least for my device, type always returns unknown

that's a libratbag bug, probably because no-one needed it so the implementation is likely missing. but I would cater for this in a way that: if the type is not unknown, use the type, otherwise use the button number.

> Is there some way to put more descriptive actions here, such as "left click"? 

I think this is something where the GUI has to make the decisions and limit what can be done compared to the raw interface on the commandline. It's acceptable to have a list of "left click", "right click", ... etc and then map those to the respective button numbers without having to expose the button number option to the user. The wacom configuration is a guideline here, see https://wiki.gnome.org/Design/SystemSettings/Tablet though that one is not up to date. Look at the stylus config, it says "Left button click". I think we should do the same here.

Come to think of it, the GtkStack might be overkill as-is. To the user Button mapping and Key mapping are effectively the same, you can have a 'Mapping' stack page and a 'Macro' stack page. The mapping one has a combo box for the options from left click through to "Send Keystroke" which would show an additional element like the current wacom config does. The others are simple entries.

Macro is Key Sequence and Custom combined, @bentiss and I talked about this ages ago and iirc we decided to drop the Key Sequence because it's effectively the same as a macro anyway.

And now that I thought of the first question here again, if we go with the above we can have the actual action listed as button label.  The visual guides are sufficient for identifying the button anyway. So we can list "Left click", "Key stroke", "Macro", etc. instead of the button number

Hjdskes:
> that's a libratbag bug, probably because no-one needed it so the implementation is likely missing. but I would cater for this in a way that: if the type is not unknown, use the type, otherwise use the button number.

I opened https://github.com/libratbag/libratbag/issues/233 for that; I'll just assume it's fixed in Piper and implement your suggestion.

> It's acceptable to have a list of "left click", "right click", ... etc and then map those to the respective button numbers without having to expose the button number option to the user. The wacom configuration is a guideline here, see https://wiki.gnome.org/Design/SystemSettings/Tablet though that one is not up to date. Look at the stylus config, it says "Left button click". I think we should do the same here.

That sounds good, but what do I do with thumb buttons? In libratbag, these are simply named `thumb`, `thumb2`, `thumb3`, etc... but I'll need to differentiate them somehow for the users. The same goes for other types of buttons such as those on top.

> Come to think of it, the GtkStack might be overkill as-is. To the user Button mapping and Key mapping are effectively the same, you can have a 'Mapping' stack page and a 'Macro' stack page. The mapping one has a combo box for the options from left click through to "Send Keystroke" which would show an additional element like the current wacom config does. The others are simple entries.

I'll play around with this idea, sounds like something that should be doable.

> Macro is Key Sequence and Custom combined, @bentiss and I talked about this ages ago and iirc we decided to drop the Key Sequence because it's effectively the same as a macro anyway.

Until this is implemented in libratbag I'll just go with what is currently there, it shouldn't be hard to adjust once libratbag makes this change.

> And now that I thought of the first question here again, if we go with the above we can have the actual action listed as button label. The visual guides are sufficient for identifying the button anyway. So we can list "Left click", "Key stroke", "Macro", etc. instead of the button number

This has been the plan all along :)

whot:
for the archives: @Hjdskes was talking about the button *configuration* page, I was talking about the main buttons page where the button to configure is selected (same level as LED/Resolution pages). Hence the slight confusion

cvuchener:
If I understand correctly the buttons you want to rename are the *functions* you are remapping the physical buttons to. I don't think you should rename them according to the physical location of the button on a particular device. What is important when choosing the remapping is what the applications will receive: `BTN_MOUSE+x` events, some of them have particular roles, others are completely generic.

In addition, the number of buttons you can remap to is different from the number of actual buttons on the device. The G500 is a 10 button mouse but it can be remapped to buttons 1 to 16, and buttons 7 and 8 (the wheel tilt buttons) default mapping is not to generic buttons, so remapping them to button 7 or 8 (generic buttons) is not the same as remapping them to wheel tilt left or right (horizontal wheel).

I think you should rename only the first three or five buttons (Left, Right, Middle, Back, Forward), the rest are generic button functions (Button 6, Button 7, ...). I am not sure exactly what are the standard roles, but that is the idea.

piper/buttondialog.py
```diff
@@ -102,13 +128,112 @@ def _init_mapping_page(self, buttons):
...
+
+    def do_key_press_event(self, event):
+        # Overrides Gtk.Widget's standard key press event callback, so we can
+        # capture the pressed buttons in capture mode. Gratefully copied from
+        # GNOME Control Center's keyboard panel.
+        # Don't process key events when we're not in capture mode.
+        if self.stack_mapping.get_visible_child_name() == "overview":
+            return Gtk.Widget.do_key_press_event(self, event)
+
+        # TODO: remove this workaround when libratbag removes its keycode
```
Hjdskes: I've already spoken with @whot about this. His solution is to remove the restriction of the key mapping signature (e.g. a single regular keycode followed by zero or more modifier keycodes) from libratbag, so that we don't have to make the distinction between regular keys and modifiers anymore (this will, I think, significantly clean up both this callback and the `Keystroke` class (see below), so I'd like it to happen sooner rather than later). It will likely take a while to do this so this is the current workaround. Not really looking for feedback here (just pointing out this issue, mainly for @whot) but feel free to comment with an alternative solution. Full problem description below:

The problem is that we can't rely on GdkEventKey's `state` member because
  1. it is 0 for individual modifier key presses (it contains the modifiers that apply before the key press happens), and
  2. we can't listen only for regular keycodes and apply their `state` mask because we cannot map from state mask to individual key codes (e.g., we can detect that control was pressed but not if it was left or right control) which we need.

So, the solution is to capture individual key presses as they occur and check GdkEventKey's `is_modifier` member instead to check if they are a modifier key. This member is however not exposed through PyGObject (https://bugzilla.gnome.org/show_bug.cgi?id=752784) and hence we need to approximate its behavior.

piper/buttondialog.py
```diff
@@ -102,13 +128,112 @@ def _init_mapping_page(self, buttons):
...
+        if keyval_lower == Gdk.KEY_Sys_Req and (real_mask & Gdk.ModifierType.MOD1_MASK):
+            keyval_lower = Gdk.KEY_Print
+
+        # Backspace disables the current shortcut.
+        if not event.is_modifier and real_mask == 0 and keyval_lower == Gdk.KEY_BackSpace:
+            self._keystroke.clear()
+            self._set_keystroke()
+            return Gdk.EVENT_STOP
+
+        if event.is_modifier:
```
Hjdskes: As said above, this if/else can go once libratbag doesn't care about its key mapping signature anymore.

piper/keystroke.py
```diff
@@ -0,0 +1,128 @@
...
+        return self._accel
+
+    def set_key(self, keycode, keyval):
+        """Sets the given keycode and keyval as the key of this keystroke.
+
+        @param keycode The raw key code as reported by GDK, as int.
+        @param keyval The key, as int. See the `gdk/gdkkeysyms.h` header file
+                      for a complete list of GDK key codes.
+        """
+        # TODO: is a validity check necessary here? In principle, all uses of
```
Hjdskes: Here is where I'd like some feedback (the TODO below is the same)

piper/keystroke.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def from_evdev(self, keycode, modifiers):
+        """Converts the given evdev keycodes to their GDK counterparts and sets
+        them as the current keystroke.
+
+        @param keycode The evdev keycode, as int.
+        @param modifiers A list of modifier keycodes, as [int].
+        """
+        # TODO: get keyval from keycode. Without this, the accelerator won't be
+        # printed correctly.
```
Hjdskes: Here too I'd like some feedback. I've been looking into this for a few hours now and there doesn't seem to be a way to go from keycode to keyval (Gdk, not evdev) that is 100% accurate.

piper/keystroke.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def to_evdev(self):
+        """Returns a list of keycodes with the first being the pressed key and
+        the rest modifiers, if any. These keycodes map to those as defined by
+        evdev as opposed to those defined by GDK, thus the returned list is
+        directly usable by libratbag."""
+        if self._keycode > 0:
+            ret = [self._keycode - self._XORG_KEYCODE_OFFSET]
+        else:
+            # TODO: this gives KEY_RESERVED; what does libratbag do with that?
```
Hjdskes: Another one: I have no idea what will happen if we set a KEY_RESERVED.

piper/keystroke.py
```diff
@@ -0,0 +1,128 @@
...
+            ret = [self._keycode]
+        return ret + [keycode - self._XORG_KEYCODE_OFFSET for keycode, _ in self._modifiers]
+
+    def _update_accel(self):
+        # Updates the accel property, so that any GObject bindings to
+        # Gtk.ShortcutLabels' accelerator property are updated automatically.
+        mask = Gdk.ModifierType(0)
+        for (_, keyval) in self._modifiers:
+            # The following are GTK+'s default modifiers, hence we check for
+            # these modifiers only -- it is unlikely we will find others.
```
Hjdskes: This is basically the same issue as with GdkEventKey's `state` member, but reversed. Again I'm approximating its behavior here, but it should be good enough.

Hjdskes:
After working on key mappings for a few days I've come to a point where I've hit a few issues I cannot resolve by myself; I've commented them inline. @whot @bentiss or anyone else (feel free to CC) please take a look and advice if you can.

piper/buttondialog.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@GtkTemplate(ui="/org/freedesktop/Piper/ui/ButtonDialog.ui")
+class ButtonDialog(Gtk.Dialog):
+    """A Gtk.Dialog subclass to implement the dialog that shows the
+    configuration options for button mappings."""
+
+    __gtype_name__ = "ButtonDialog"
+
+    _BUTTON_TYPE_TO_DESC = {
```
Hjdskes: I'm thinking of moving this to ratbagd.py; I'm going to have to add another dictionary like this for the special actions. Ideally, both of these will be in the bindings so they can more easily be kept in sync with libratbag.

Hjdskes:
@whot Feel free to try and comment on the special mappings page; that one is done.

As for the UI, I'm thinking I can merge the special mapping combobox with the button mapping combobox, and perhaps I can get the key mapping button in there as well.

Otherwise we can maybe present a large listbox like GNOME Control Center's keyboard page does, which then also includes an entry for macros (and old macros that you can remember but that aren't currently assigned?)

piper/buttondialog.py
whot: yeah, that seems like a good idea. note that I have a FIXME in my dbus rework branch - having the types exposed as strings doesnt make sense, we use enums everywhere else. So expect that to change soon too.

piper/buttondialog.py
whot: A few things to note here, @bentiss can correct me if I misremember: libratbag doesn't actually care about modifiers, I think the only reason why we handle them is so we can order the keys correctly. i.e. "push all modifiers down, key down/up, modifiers up".  That's about it. And it's not even hooked up in libratbag, there's just a fixme there.

So I think you can already make this code a lot simpler by just going the same path: catch the events as they occur and assume that the last key pressed was the real key and all before that are still down are modifiers. It just doesn't matter otherwise.

Or, come to think of it, it may be easier to just leave this bit out and only allow button assignments. Macros are something we already support on some devices, it'd be better to just work on the macros.

piper/keystroke.py
whot: I'm not sure how you'd do a validity check here tbh. keycode to keyval conversion is notoriously difficult and we really want to stay out of that. It's relatively easy to go code->val with XKB (that's what it's designed for) but it's hard and sometimes impossible to go val->code. So any validation is largely pointless IMO.

piper/keystroke.py
whot: you have to go through XKB but tbh I don't know exactly how this is handled here.  `gdk_keymap_translate_keyboard_state ()` seems like it's a good start, but you'll run into the various modifier issues here. 

I think the best option here may be to use `gdk_keymap_get_entries_for_keycode ()` and always select the entry on level 0. That gives you a key you can then feed into `gdk_keymap_lookup_key ()` to get the keyval, which has a printable name with `gdk_keyval_name()`

Together with a few warnings that we're recording keys not *syms* this may be enough (at least for now). It just means that when you're recording "shift 1" you don't record "!" but really "shift 1". At least using the get_entries call should provide basic layout mapping, so we don't have a confusion when someone uses azerty.

piper/keystroke.py
```diff
@@ -0,0 +1,128 @@
...
+            keyval = 0
+            self.add_modifier(keycode, keyval)
+
+    def to_evdev(self):
+        """Returns a list of keycodes with the first being the pressed key and
+        the rest modifiers, if any. These keycodes map to those as defined by
+        evdev as opposed to those defined by GDK, thus the returned list is
+        directly usable by libratbag."""
+        if self._keycode > 0:
+            ret = [self._keycode - self._XORG_KEYCODE_OFFSET]
```
whot: shouldn't this be `self._keycode > self._XORG_KEYCODE_OFFSET`?

piper/keystroke.py
whot: it's an unmapable key. Ignore it and warn about it in the gui?

piper/keystroke.py
Hjdskes: I didnt' necessarily mean verifying that the keycode belongs to the keyval, but more that both are valid integers. For example, `passing -1` here for the keycode (for whatever reason), would make `Gtk.accelerator_name(keyval, mask)` error out. So what we could do here is to check whether `keycode` and `keyval` are within a certain range, e.g. for `keyval` this could be `Gdk.KEY_BackSpace <= keyval <= Gdk.KEY_LogGrabInfo`. Though I don't think that the `keyval` values are necessarily consecutive so this might give a false sense of security, and I'm not sure what these values would be for `keycode`.

piper/keystroke.py
Hjdskes: This would happen because it's our default keycode value, so if the user attempts to set an initially emptry keystroke but then bails out with BackSpace, this method will be called on a keycode of value 0.

piper/keystroke.py
Hjdskes: Yes, good catch :)

piper/buttondialog.py
Hjdskes: Strings are actually working very nicely for me right now but I can make enums work too. Also, I could just place the names in the .ui file so no need for that second dictionary. Still, probably best to move that kind of stuff to the bindings.

whot:
At least the key codes would be between 8 and 255 because X can't
route anything above 255. So that's an easy one to check for. Wayland can
route those and it doesn't have the offset but as I said we can worry about
that later. I don't think you need to do anything else beyond this yet.

piper/keystroke.py
Hjdskes: I think I tried all of these but as you say it's impossible to be 100% accurate, but I suppose providing a hint/warning that we're recording keys instead of syms could work. I'm not sure if there are any applications/game engines/what have you that won't work with just keycodes? If not, I suppose this is a valid workaround for the problem.

You say I always select the entry on level 0; what about the `group`? I assume also 0, but just checking to be sure.

whot:
in that case it's even easier to silently ignore? :)

piper/buttondialog.py
Hjdskes: > libratbag doesn't actually care about modifiers, I think the only reason why we handle them is so we can order the keys correctly. i.e. "push all modifiers down, key down/up, modifiers up". That's about it. And it's not even hooked up in libratbag, there's just a fixme there.

Sure, but you still need to know whether left or right Alt/Control/Shift/whatever is pressed, don't you?

> So I think you can already make this code a lot simpler by just going the same path: catch the events as they occur and assume that the last key pressed was the real key and all before that are still down are modifiers. It just doesn't matter otherwise.

"That are still down"; that'd be the `state` mask I'm talking about. Are you saying I need to cache keypresses and for every new key press, check the cached ones against the state mask and throw away anything that isn't set in the mask? That'd be tricky to get right, but I can try..

> Or, come to think of it, it may be easier to just leave this bit out and only allow button assignments. Macros are something we already support on some devices, it'd be better to just work on the macros.

How exactly do macros differ from button- and key mappings? It seems like there isn't really a difference here so either way this problem will have to be solved, be it for key mappings or macros.

piper/keystroke.py
whot: At least the key codes would be between 8 and 255 because X can't route anything above 255. So that's an easy one to check for. Wayland can route those and it doesn't have the offset but as I said we can worry about that later. I don't think you need to do anything else beyond this yet.

piper/keystroke.py
whot: in that case it's even easier to silently ignore? :)

piper/keystroke.py
whot: in XKB lingo: "level" == shift state, so your normal key has two levels 'a' and 'A'. it gets more complicated after that. "group" == what you'd think of as layout. Ideally you choose the one currently active, but if you don't 0 is a good enough option. It'd only be an issue when users have more than one layout configured and switch between them. So if gdk has something to get the group use that, otherwise pick 0.

piper/buttondialog.py
whot: sorry, last 30s comment before I run off: you keep your own state of down/up and replay that in the same sequence. if you never saw a key up, that key is still down.

piper/keystroke.py
Hjdskes: Alright, I've gone with `gdk_keymap_get_entries_for_keycode ()` and both level and group `0` (there is no way to determine an active keyboard group). I haven't yet added a message/warning that we're recording keys and not syms, because I'm not sure where best to put that (if at all). You should only notice anything being off when opening the dialog and an existing key mapping is displayed.

piper/keystroke.py
Hjdskes: Not sure what you meant here and we never settled on someting on IRC, but if you mean we return some kind of value that signals ratbagd to *unset* the key mapping we need to think about how to implement that. Perhaps such a reserved key can be the signal, or the empty list. @bentiss I would appreciate any input from you here as well :)

Hjdskes:
I just amended some changes that make a few things work better.

Firstly, I moved some key handling things into `KeyStroke` itself, so all that the button dialog now does is fix the `GdkEventKey` and check for BackSpace (I still want to put Escape back in to abort any changes made as well).

Secondly, in addition to fixing `is_modifier`, the key press handler in `ButtonDialog` now also masks out any modifiers other than the defaults reported by [`gtk_accelerator_get_default_mod_mask ()`](https://developer.gnome.org/gtk3/stable/gtk3-Keyboard-Accelerators.html#gtk-accelerator-get-default-mod-mask) (see `gtk_accelerator_set_default_mod_mask ()` above it for documentation on what those defaults are). This makes the checks that fix `is_modifier` and `KeyStroke._update_accelerator()` more robust.

Thirdly, as mentioned in [my comment](https://github.com/libratbag/piper/pull/47#discussion_r128742592), I've gone with `gdk_keymap_get_entries_for_keycode ()` and both level and group `0` to map Gdk keycodes to Gdk keyvals. I haven't yet added a message/warning that we're recording keys and not syms, because I'm not sure where best to put that (if at all). You should only notice anything being off when opening the dialog and an existing key mapping is displayed.

Finally, I've added a crude keycode validation method that might not be fully working with Wayland (it does in my basic tests here so far):

```python
def _check_key(self, keycode, keyval):
    # Checks if the given keycode and keyval are valid through a simply
    # range check.
    if not self._XORG_KEYCODE_OFFSET <= keycode <= 255:
        print("Keycode is not within the valid range.", file=sys.stderr)
        return False
    elif not Gdk.KEY_space <= keyval <= Gdk.KEY_AudioMicMute:
        print("Keyval is not within the valid range.", file=sys.stderr)
        return False
    return True
```

piper/buttondialog.py
```diff
@@ -0,0 +1,300 @@
...
+
+from .gi_composites import GtkTemplate
+from .keystroke import KeyStroke
+
+import gi
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gdk, GObject, Gtk
+
+
+class MappingType(IntEnum):
```
Hjdskes: This should move to the bindings.

piper/buttondialog.py
```diff
@@ -0,0 +1,300 @@
...
+        "resolution-cycle-up": _("Resolution cycle up"),
+        "resolution-up": _("Resolution up"),
+        "resolution-down": _("Resolution down"),
+        "profile-cycle-up": _("Profile cycle up"),
+        "profile-up": _("Profile cycle up"),
+        "profile-down": _("Profile cycle down"),
+    }
+
+    _SPECIAL_TYPES = {
+        "none": _("None"),
```
Hjdskes: This should move to the bindings.

Hjdskes:
With the last few commits that replace key mappings with macros, I think I have fixed most of the pain points discussed earlier. Because there is currently no support for macros in ratbagd, I have not yet been able to test this properly. Also, support is currently limited to key presses only; key releases can be added rather easily, key press *and* release is somewhat more difficult as Gtk does not have the combined concept. It will probably require checking if a key press and related release happened within a certain interval; not sure if that's possible in a nice way.

AFAIK the only part left here now is to display the correct label in the `ButtonsPage` itself, which I'll work on next. Feel free to test what's there meanwhile.

bentiss:
> AFAIK the only part left here now is to display the correct label in the ButtonsPage itself, which I'll work on next.

FWIW, this is something also required for ratbagctl (it's on my TODO list too). IIRC @whot told me that there are some evemu or libevdev python bindings that have the matching between name and int value that we can use.

Hjdskes:
> FWIW, this is something also required for ratbagctl (it's on my TODO list too). IIRC @whot told me that there are some evemu or libevdev python bindings that have the matching between name and int value that we can use.

That is something else and something I've already done :) Check the code in `KeyStroke.py`, and https://python-evdev.readthedocs.io/en/latest/

Hjdskes:
That part is now also done, but due to https://github.com/libratbag/libratbag/issues/233 and ratbagd issues (`GLib.Error: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.PropertyReadOnly: Property 'ButtonMapping' is not writable. (44)`) I cannot test this one either.

bentiss:
Sorry I read too fast your previous comment.

Regarding https://github.com/libratbag/libratbag/issues/233 I think you mixed up the `Type` property and the `ActionType` property, which is updated when you call the correct Set*Mapping. With ratbagctl I can update buttons, so I guess piper should also be capable of the same.

> That is something else and something I've already done :) Check the code in `KeyStroke.py`

I would think this should be part of ratbagd.py. This PR adds the translation for strings to specials so I guess it would be better to not use integer for any key and this should be masked in the binding.

Hjdskes:
I commented in https://github.com/libratbag/libratbag/issues/233 to keep that discussion in one place.

> I would think this should be part of ratbagd.py. This PR adds the translation for strings to specials so I guess it would be better to not use integer for any key and this should be masked in the binding.

I'm sorry but I'm not sure what you mean. @whot is planning on encoding the different special functions (`SpecialMapping`) with an enum, but that isn't related to mapping from `linux/input.h` keycodes to key names.

whot:
>  key press and release is somewhat more difficult
not needed, just handle them as separate press/release events. libratbag doesn't have a concept of key type anyway, it was all hidden by the ratbag-command UI

> There is no support for setting timeouts yet (how would one do that from Piper?)
I'd say a checkbox for "include time in macro" in the macro box. so no specific timeouts, but piper can record the time between the key events as the user supplies them.

> the correct label in the ButtonsPage itself
I think the button type support should be ready to use with https://github.com/libratbag/libratbag/pull/248

> I would think this should be part of ratbagd.py.
Not sure the string types should be part of ratbagd.py. I'd rather keep that one simple and leave complex things up to the real UIs

Hjdskes:
> not needed, just handle them as separate press/release events.

I can do that, but would rather in a new PR as this one is getting quite large.

> I'd say a checkbox for "include time in macro" in the macro box. so no specific timeouts, but piper can record the time between the key events as the user supplies them.

Same as above, albeit a bit more tricky. Definitely something I want to look into, but probably as a feature on its own.

> I think the button type support should be ready to use with https://github.com/libratbag/libratbag/pull/248

@bentiss explained to me yesterday that button type isn't the way to go, see https://github.com/libratbag/libratbag/issues/233

> I'd rather keep that one simple and leave complex things up to the real UIs

Indeed this is why I did it in Piper.

Hjdskes:
Hm... rebasing things in a separate branch still managed to "infect" this PR with GitHub's references...

Anyway, I just rebased this branch on `rewrite`, which in turn is rebased on current `master`. In the process I significantly cleaned up the history and the commits themselves, which due to the rebase were messier than they need to be (for example when I initially wrote this code, button action types were still strings).

As far as I am concerned this PR is now done and ready for testing and full review. Any remaining features have been turned into issues (see #56 and #57); if there are more please make an issue for those so that this large PR can at least be merged. Fixes to the PR itself will of course be done here :)

whot:
you need to rebase this on top of  libratbag commit 7c313a31fd7e9e7c8a7bb7365de58156574f856c because I'm getting all sorts of crashes for the special mapping on my resolution button, sorry.

piper/ratbagd.py
```diff
@@ -490,6 +490,26 @@ class RatbagdButton(_RatbagdDBus):
...
+    ACTION_SPECIAL_RESOLUTION_DOWN = (1 << 30) + 10,
+    ACTION_SPECIAL_RESOLUTION_ALTERNATE = (1 << 30) + 11,
+    ACTION_SPECIAL_RESOLUTION_DEFAULT = (1 << 30) + 12,
+    ACTION_SPECIAL_PROFILE_CYCLE_UP = (1 << 30) + 13,
+    ACTION_SPECIAL_PROFILE_CYCLE_DOWN = (1 << 30) + 14,
+    ACTION_SPECIAL_PROFILE_UP = (1 << 30) + 15,
+    ACTION_SPECIAL_PROFILE_DOWN = (1 << 30) + 16,
+    ACTION_SPECIAL_SECOND_MODE = (1 << 30) + 17,
+    ACTION_SPECIAL_BATTERY_LEVEL = (1 << 30) + 18,
+
```
whot: this bit has to go, it removes the fixes from benjamin's commit in libratbag, with the commas in the end the entries end up as tuples

Hjdskes:
Sorry about that, missed that during the rebasing. Everything should be fine now. If things still crash on special mappings, make sure your device has the new settings uploaded ("new" being since the DBus interface rewrite) by setting a new special mapping through ratbagctl.

whot:
huh, found a strange bug here: one button defaults to "backward", if I change it to "forward", Apply, it displays correctly. click on the button again, "Backward" is highlighted by default instead of forward. Clicking Apply doesn't do anything, I have to click on Backward (despite it being highlighted) and then click apply. Something is cached somewhere.

Hjdskes:
After some more testing I found a few more strange things (on libratbag master):

1. The one reported above by @whot.
2. Special mappings are not displayed in the configuration stack page, even though they are set on the device.
3. Macros are not being applied to my device.
4. Macros are not displayed in the configuration stack page.

Looking into these now...

bentiss:
> 3. Macros are not being applied to my device.

yeah, we know :)
https://github.com/libratbag/libratbag/blob/master/src/driver-hidpp20.c#L364

It's on my TODO list

Hjdskes:
Good to know, one less to look into. FWIW the others seem to be related to the same issue, which appears to be a race condition in the bindings again.

Hjdskes:
After https://github.com/libratbag/libratbag/pull/264, 
27062a2b59e3f9846d568bac7b29a97c69bf32ab and some amending to other commits here I believe every issue noted above is now fixed. I haven't crossed off number four as I cannot actually test that for my device, due to its driver not supporting macros yet.

@whot Please review again, when you have time :)

PR #51, opened 2017-07-24
ratbagd: remove key mappings and add macros

Hjdskes:
Since Piper will always use macros for both key mappings and, well, macros, there is no need to keep the key mappings around. Eventually these will both be merged in ratbag as well (see https://github.com/libratbag/libratbag/issues/243).

Note that this isn't yet added in ratbagd and as such this code does not yet work. For that reason, I based it on `rewrite` and not master.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -420,6 +420,9 @@ class RatbagdResolution(_RatbagdDBus):
 class RatbagdButton(_RatbagdDBus):
     """Represents a ratbagd button."""
 
+    KEY_PRESS = 0
+    KEY_RELEASE = 1
+
     def __init__(self, object_path):
         _RatbagdDBus.__init__(self, "Button", object_path)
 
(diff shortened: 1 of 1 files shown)


whot:
looks good, but obviously needs the ratbagd bits first before we start using the bindings. keep that in your tree, I'll try to get the ratbagd side asap

Hjdskes:
It also needs to be merged with the new ratbagd.py from libratbag.

piper/ratbagd.py
```diff
@@ -420,6 +420,9 @@ def set_default(self):
 class RatbagdButton(_RatbagdDBus):
     """Represents a ratbagd button."""
 
+    KEY_PRESS = 0
+    KEY_RELEASE = 1
```
Hjdskes: Should add `KEY_PRESS_RELEASE` here as well, even though Piper for now will have only key press support.

whot: actually, no. libratbag doesn't use that, if you press+release, that is two events in the macro. this would have to be abstracted in the next layer, imo.

Hjdskes:
Closing this as this is part of https://github.com/libratbag/libratbag/pull/248, minus the removing of key mappings. That's probably better in https://github.com/libratbag/libratbag/pull/243 anyway.

PR #52, opened 2017-07-24
Make Piper rewrite work with the new bindings

Hjdskes:
This code is most likely temporarily until I get around to implementing profiles for real.

### DIFF
--- a/piper/ledspage.py
+++ b/piper/ledspage.py
@@ -42,7 +42,7 @@ class LedsPage(Gtk.Box):
         self._init_ui()
 
     def _init_ui(self):
-        profile = self._device.active_profile
+        profile = self._find_active_profile()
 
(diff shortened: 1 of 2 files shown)


piper/ledspage.py
```diff
@@ -57,6 +57,12 @@ def _init_ui(self):
             mousemap.add(button, "#led{}".format(index))
             sizegroup.add_widget(button)
 
+    def _find_active_profile(self):
+        # Finds the active profile, which is guaranteed to be found.
+        for profile in self._device.profiles:
+            if profile.is_active:
+                return profile
```
whot: I was googling around and found this construct, figured I'd share:
````python
return next(p for p in self._device.profiles if p.is_active)
````
Also takes a default parameter, so could be used with 
````python
return next((p for p in self._device.profiles if p.is_active), self._device.profiles[0])
````
Just sharing this, I hadn't known about it :)

PR #54, opened 2017-07-24
RatbagdDBusUnavailable: add exception message

Hjdskes:
Quite a few PRs open both here and in libratbag that now modify the bindings. We should be careful when merging all of them.

In any case, this adds exception messages to `RatbagdDBusUnavailable` so that we can print those to the user.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -54,17 +54,17 @@ class RatbagErrorCode(IntEnum):
     RATBAG_ERROR_IMPLEMENTATION = -1004
 
 
-class RatbagdDBusUnavailable(BaseException):
+class RatbagdDBusUnavailable(Exception):
     """Signals DBus is unavailable or the ratbagd daemon is not available."""
     pass
 
 
-class RatbagdDBusTimeout(BaseException):
+class RatbagdDBusTimeout(Exception):
     """Signals that a timeout occurred during a DBus method call."""
     pass
 
 
-class RatbagError(BaseException):
+class RatbagError(Exception):
     """A common base exception to catch any ratbag exception."""
     pass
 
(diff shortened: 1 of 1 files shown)


piper/ratbagd.py
```diff
@@ -54,17 +54,22 @@ class RatbagErrorCode(IntEnum):
...
-class RatbagdDBusUnavailable(BaseException):
+class RatbagdDBusUnavailable(Exception):
     """Signals DBus is unavailable or the ratbagd daemon is not available."""
-    pass
+
+    def __init__(self, value):
+        self.value = value
+
+    def __str__(self):
+        return repr(self.value)
```
whot: is this needed? doesn't Exception implement this already?

````python
>>> e = Exception('foo')
>>> str(e)
'foo'
>>> class Banana(Exception):
...     pass
... 
>>> b = Banana('foo')
>>> str(b)
'foo'

````

File: piper/ratbagd.py [hunk 2]
```diff
@@ -137,11 +139,11 @@ def __init__(self, interface, object_path):
...
                                                  self._interface,
                                                  None)
-        except GLib.Error:
-            raise RatbagdDBusUnavailable()
+        except GLib.Error as e:
+            raise RatbagdDBusUnavailable(e.message)
 
         if self._proxy.get_name_owner() is None:
-            raise RatbagdDBusUnavailable()
+            raise RatbagdDBusUnavailable("Noone currently owns {}".format(ratbag1))
```
whot: typo: 'no one'

piper/ratbagd.py
Hjdskes: Whoops, you're right. Amended my commit.

PR #58, opened 2017-07-25
Add a piper.devel output to run from source tree

### LINKED ISSUE #50: Make it possible to run piper from the source tree
`./build/piper` should just work without any further setup. Since that requires some extra bits to load the module path or at least set up the path, might be easier to just provide a `./build/piper.devel` script that mangles everything into place as required.

Hjdskes:
Fixes #50, by adding a `piper.devel` output next to the regular `piper` output. The regular `piper` output now doesn't have any debug or development code whatsoever, and the `piper.devel` output has two lines of code that 1) fix the Python path to import from the source directory and 2) print to the user that they're running from source.

I went for the approach that adds / moves around the least amount of files. It might be a bit abusive of Meson but I think overall the approach is quite minimal and the achieved effect is nice.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,7 @@ jobs:
           command: |
             dnf install -y git python3-flake8
       - checkout
-      - run: flake8-3 . piper.in
+      - run: flake8-3 .
 
       # now install and checks the resulting file (just in case)
       - run:
@@ -19,6 +19,6 @@ jobs:
           command: |
             meson build
             ninja -C build
-      - run: flake8-3 buildir/piper
+      - run: flake8-3 buil/piper buil/piper.devel
       - store_artifacts:
           path: ~/piper/build/meson-logs
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 ignore = E402,E501
-exclude = .git,__pycache__,build,data,piper/piper.py
+exclude = .git,__pycache__,build,data,piper/piper.py,piper.in
--- a/meson.build
+++ b/meson.build
@@ -24,13 +24,25 @@ if not python_dir.endswith('site-packages')
 endif
 install_subdir('piper', install_dir: python_dir)
 
-conf = configuration_data()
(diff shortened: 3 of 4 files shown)


Hjdskes:
Requested a review from @bentiss too since @whot is probably in the air right now.

Hjdskes:
> I now wonder how piper.in is generated given that it has its own history in git.

Sorry, bad choice of words here. By generated I mean that `piper` and `piper.devel` are generated from it, as in the `.in` marks that it's not a Python source file.

I suppose we can write a wrapper script that sets `PYTHONPATH`, but considering that `piper.in` is by now mostly static I'm not sure if that's worth it. Do we really want an entirely new source file over two lines of inserted code?

You decide :)

bentiss:
> Do we really want an entirely new source file over two lines of inserted code?

This is something to take into account, but I generally don't like to have exceptions based on "oh, this file is pretty much done, so it won't change in the future".

Anyway, the good point which makes me agree with you is that circle.yml will run flake8 on the generated file, so at least we will know, when the PR is out if flake8 complains.
Talking about CircleCI, this PR should also amend the circle.yml file to remove the flake8 check on `piper.in` and also add a check on `piper.devel`.

Hjdskes:
> but I generally don't like to have exceptions based on "oh, this file is pretty much done, so it won't change in the future".

That's a good point.

> Talking about CircleCI, this PR should also amend the circle.yml file to remove the flake8 check on piper.in and also add a check on piper.devel.

I've amended the last commit to include these changes.

circle.yml
```diff
@@ -19,6 +19,6 @@ jobs:
           command: |
             meson build
             ninja -C build
-      - run: flake8-3 buildir/piper
+      - run: flake8-3 buildir/piper buildir/piper.devel
```
Hjdskes: Shouldn't this be `builddir` (notice the extra `d`)?

bentiss: oops, yes :)

Hjdskes: Oke, amended agan.

PR #63, opened 2017-07-27
Merge rewritten Piper into master

Hjdskes:
With the latest changes to the DBus interface, Piper master is broken even though we tried to be careful. As it makes no sense to do twice the work, and the rewritten Piper almost supports the same features as the old Piper, we merge the `rewrite` branch now.


whot:
sorry, should've merged this with a proper pull, I quickly force-pushed to master and merged it again with `--no-ff` so we have the history show the pull

PR #64, opened 2017-07-27
ratbagd: cache lists to ensure we return the same objects

Hjdskes:


piper/ratbagd.py
```diff
@@ -206,14 +206,15 @@ class Ratbagd(_RatbagdDBus):
...
+
     @GObject.Property
     def devices(self):
         """A list of RatbagdDevice objects supported by ratbagd."""
-        devices = []
-        result = self._get_dbus_property("Devices")
-        if result is not None:
-            devices = [RatbagdDevice(objpath) for objpath in result]
-        return devices
+        return self._devices
```
whot: hmm, I'm fine with the rest of the changes, but this one we need to really fix. (un)plugging a device while ratbagd is running *must* work.

Hjdskes: You're right. I'll look into it soon.

Hjdskes:
I just amended the commit to make it work with device (un)plugging, where the new device is just appended to the list:

```
 > ./build/piper.devel
      /org/freedesktop/ratbag1/device/event13 removed
Done: []
      /org/freedesktop/ratbag1/device/event13 added
Done: ['/org/freedesktop/ratbag1/device/event13']

```

I also cherry-picked https://github.com/libratbag/piper/commit/e9ea7ddca77807af191464029367deb9e8a96dda from #36.

piper/ratbagd.py
```diff
@@ -205,15 +219,22 @@ class Ratbagd(_RatbagdDBus):
 
     def __init__(self):
         _RatbagdDBus.__init__(self, "Manager", None)
+        result = self._get_dbus_property("Devices")
+        self._devices = [RatbagdDevice(objpath) for objpath in result]
+
+    def _on_properties_changed(self, proxy, changed_props, invalidated_props):
+        if "Devices" in changed_props.keys():
+            for prop in changed_props["Devices"]:
+                index = self._find_object_with_path(self._devices, prop)
```
bentiss: Naive question: if you implement a `__eq__` method on the RatbagdDevices, couln't you just replace these by:
```
if prop in self.devices:
    self._devices.remove(prop)
else:
    self._devices.append(prop)
```
?

whot: good point, yes, that should work

Hjdskes: I'll update this sunday :)

Hjdskes: Looking at this some more, I don't think it's that simple. `prop` is a string (the object path), so before you can call `remove` or `append` you still need to find the `RatbagdDevice` with the right object path.

whot: that's what `list.remove()` does for you though, it calls (iirc) `__eq__` on each device in the list. So if `__eq__` is implemented to return true if `self.objpath == that_string`, then it should work.

it's not pretty because now you have to implement `__eq__` when comparing against an object and a string, but it's doable. whether it's good python I'm not sure. Feel free to leave it as-is, `__eq__` doesn't give us that much benefit here.

Hjdskes: `So if __eq__ is implemented to return true if self.objpath == that_string, then it should work.`

That will work yes, but I'm not sure about it either. Let's leave this as-is for now; the removal is a bit more lines of code than it can be but at least the complexity is in an obvious place; I don't like hiding it away in `__eq__`.

PR #73, opened 2017-07-31
Implement profiles

Hjdskes:
This PR adds support for profiles, i.e. it allows the user to switch between profiles, disable them (which hides them from the UI) and enable them (make the first disabled profile visible again).

@whot Just a WIP PR to get an ACK on the current approach before I extend it to everything.

### DIFF
--- a/piper/profilerow.py
+++ b/piper/profilerow.py
@@ -1,0 +1,58 @@
+# Copyright (C) 2017 Jente Hidskes <hjdskes@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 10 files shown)


piper/profilerow.py
```diff
@@ -0,0 +1,54 @@
...
+
+    title = GtkTemplate.Child()
+
+    def __init__(self, profile, *args, **kwargs):
+        Gtk.ListBoxRow.__init__(self, *args, **kwargs)
+        self.init_template()
+        self._profile = profile
+        self._profile.connect("notify::enabled", self._on_profile_notify_enabled)
+
+        self.title.set_text(_("Profile {}").format(profile.index))
```
Hjdskes: Yup, will do!

Hjdskes:
Ok, I think this is ready for full review now. I do sometimes get the following, which seems to occur if the user commits any changes to the device quickly after switching profiles:

```
 > ./build/piper.devel
  File "/home/jente/code/src/github.com/libratbag/piper/piper/ratbagd.py", line 195, in _dbus_call
    500, None)
GLib.Error: g-io-error-quark: Timeout was reached (24)

During handling of the above exception, another exception occurred:

  File "/home/jente/code/src/github.com/libratbag/piper/piper/window.py", line 134, in _on_save_button_clicked
    status = self._device.commit()
  File "/home/jente/code/src/github.com/libratbag/piper/piper/ratbagd.py", line 327, in commit
    return self._dbus_call("Commit", "")
  File "/home/jente/code/src/github.com/libratbag/piper/piper/ratbagd.py", line 202, in _dbus_call
    raise RatbagdDBusTimeout(e.message)
piper.ratbagd.RatbagdDBusTimeout: Timeout was reached
```

I'm not sure what's the best way to resolve this.

PR #86, opened 2017-08-02
Application: change aboutDialog.ui resource path to AboutDialog.ui

julianrichen:
The current about dialog is named `AboutDialog.ui` but is linked as `/org/freedesktop/Piper/aboutDialog.ui` in `src/application.py`. Which throws:
```
(piper.devel:6977): Gtk-ERROR **: failed to add UI: The resource at “/org/freedesktop/Piper/aboutDialog.ui” does not exist
```

This pr fixes that.

### DIFF
--- a/piper/application.py
+++ b/piper/application.py
@@ -63,7 +63,7 @@ class Application(Gtk.Application):
 
     def _about(self, action, param):
         # Set up the about dialog.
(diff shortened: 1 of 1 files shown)


Hjdskes:
Good find, thanks!

PR #91, opened 2017-08-03
Drop indices from the GUI

Hjdskes:
The resolution rows look quite empty now; not sure what to do with those.


whot:
Too much whitespace in the resolution selection now (between 1234dpi and the x), we should try to make this narrower.

Issue #68, opened 2017-07-30
Resolution sliders need stops and reasonable ranges


whot:
Comment from @hadess during GUADEC when I demoed it: the resolution sliders should have increments and snap to those (e.g. every 50 or 100 dpi).

Also, the resolution range can be capped to a resonable range, if we don't expect anyone to set 12000 dpi, we shouldn't display it.

Hjdskes:
They do have increments at least right? Otherwise something broke; I'm sure I added this. 

Snapping I can add, if we can hide the indicators. I tried that before but there are so many steps that there indicators form a thick black bar under the scale :sweat_smile:

Hjdskes:
Just looked into this: the scales have both increments and snaps, but because the increment is so small relative to the range (50 vs 50 to 12000) the snapping is barely noticeable. The only way to resolve this is to reduce the range, but there are people who might want to use 12000...

In case you haven't noticed, long pressing the scale enlarges it and makes it less sensitive to movement; it's an undiscoverable default GTK thing but it might help here.

whot:
I guess incrementing to 100 won't make much of a difference either?
I do wonder whether there's a real use-case for 12000 or whether we should just limit it to 6000 and worry about it when people complain.

Hjdskes:
I can do both, don't really have a preference here. Just let me know which you want me to pick :)

whot:
start with the 6000 first and see if the 50 stop still makes sense. if not, try the 100 on top of the 6000 clamp. If that doesn't make sense either then we leave it as-is

hadess:
I think you might want to look into using something that's more like a logarithmic scale, so that you get loads of options at the bottom end (say, 100 to 1200) and far less in the higher end.

Changing the UI to make the slider as wide as possible would also help.

whot:
This is effectively fixed with #187

Hjdskes:
@whot Can we close this now?

whot:
yes we can!

PR #94, opened 2017-08-07
ResolutionsPage: reduce width of the controls

Hjdskes:
As mentioned in #91. I think this is the smallest we can make it while it still looks OK.

### DIFF
--- a/data/ui/ResolutionsPage.ui
+++ b/data/ui/ResolutionsPage.ui
@@ -10,7 +10,7 @@
     </child>
     <child>
       <object class="GtkBox">
-        <property name="width_request">400</property>
+        <property name="width_request">300</property>
         <property name="visible">True</property>
         <property name="can_focus">False</property>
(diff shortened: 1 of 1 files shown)


whot:
sorry to bump you around here but such is the life of UI design. As @hadess noted in #68, maybe the solution to the whole thing is to make them wider so the range is more accessible.?

whot:
uhm, on that note, maybe we should just show the sliders directly for each of them?

Hjdskes:
I'm not convinced that making them wider is the best solution; I don't think it'll make much of a difference and it's certainly not going to look better. In the same way, moving them up into the gap of whitespace that this PR is trying to remove will mean that they're even smaller than they are now, which we certainly don't want.

In #68 @hadess suggested a logarithmic scale; I'm not sure if that possible but I'd like to experiment with that to see what we can get. We can always revert to the brute "make everything as wide as possible" option later :)

PR #97, opened 2017-08-08
React to devices being added or removed

Hjdskes:
This should go after #84.

Maybe we can show an [empty placeholder](https://developer.gnome.org/gtk3/stable/GtkListBox.html#gtk-list-box-set-placeholder) in the welcome perspective's listbox when there are no devices, instead of displaying the error perspective. This would look something like the empty search result in GNOME Control Center's keyboard panel:

screenshot from 2017-08-08 17-38-45

Right now, Piper does the following:

Device added:
* No device connected currently → immediately configure the connected device
* Devices currently connected, in welcome perspective → add the new device to the list
* Device(s) currently connected, in mouse perspective → make the back button visible (will be added in this PR, currently not there), possibly an in-app notification if we want it

Device removed:
* Removed device is the one currently being edited → error perspective
* Device(s) currently connected, in welcome perspective → remove it from the list. If it was the last device, display the error perspective
* Devices currently connected, configuring another in the mouse perspective → hide the back button if required (will be added in this PR, currently not there), possibly an in-app notification if we want it

Of course you can mix and match these, for example if you are configuring a device, disconnect it and then connect it again you'll jump straight back into editing it. Please let me know if there's a scenario that I missed!

Also, I think that the Window class is getting quite messy. I'd like to clean it up a bit and get some more structure in there, but I'll see what I can do given the time frame and other, more important issues.

### DIFF
--- a/piper/application.py
+++ b/piper/application.py
@@ -50,7 +50,6 @@ class Application(Gtk.Application):
         """This function is called when the user requests a new window to be
         opened."""
         window = Window(self._ratbag, application=self)
-        window.show_all()
         window.present()
 
     def _build_app_menu(self):
--- a/piper/errorperspective.py
+++ b/piper/errorperspective.py
@@ -51,6 +51,12 @@ class ErrorPerspective(Gtk.Box):
         """The titlebar to this perspective."""
         return self._titlebar
 
+    @GObject.Property
+    def can_go_back(self):
+        """Whether this perspective wants a back button to be displayed in case
+        there is more than one connected device."""
+        return False
+
     def set_message(self, message):
         """Sets the error message.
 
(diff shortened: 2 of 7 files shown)


Hjdskes:
> Of course you can mix and match these, for example if you are configuring a device, disconnect it and then connect it again you'll jump straight back into editing it.

Forgot to mention that this doesn't work yet due to Piper crashing because of a bug ([bugzilla link](https://bugzilla.gnome.org/show_bug.cgi?id=722562)):
```
  File "/home/jente/code/src/github.com/libratbag/piper/piper/mousemap.py", line 181, in do_forall
    for child in self._children:
AttributeError: 'MouseMap' object has no attribute '_children'
```

Hjdskes:
^ Fixed with 6f61538613ef2a751bec78202f54be37e9416d6a.

whot:
sorry, this one conflicts with the current master branch now, can you rebase please? Thanks

Hjdskes:
My bad! It's rebased now :)

Issue #69, opened 2017-07-30
Save button is not obvious


whot:
Comment from @hadess when I showed him piper at GUADEC: the save button to actually write to the mouse is not obvious enough, needs to be differently placed.

Hjdskes:
Would replacing the image with simply "Save" work? I'm not sure where else we can place it to make it more obvious.

Also, in case you didn't notice, the button has a tooltip to explain its purpose.

whot:
There are a couple of comments here: first, the icon is "save to disk", that was confusion number one. Second, it's not in-your-face obvious that you have to press it to write to the device. So replacing it with a Save or "Write to device" or so is likely better. The last thing: maybe we should highlight/flash/something the button whenever the profile is dirty?

bentiss:
I still believe we should not explicitely ask the user to actually send the configuration to the mouse. I wish we could just monitor the events and automatically send the commit when there is no widget interactions for 1 second for instance

Hjdskes:
> automatically send the commit when there is no widget interactions for 1 second for instance

Two immediate problems come to mind with this approach:

1. timeouts are not discoverable
2. what happens if the user wants to abort the changes they made?

> There are a couple of comments here: first, the icon is "save to disk", that was confusion number one. Second, it's not in-your-face obvious that you have to press it to write to the device. So replacing it with a Save or "Write to device" or so is likely better. The last thing: maybe we should highlight/flash/something the button whenever the profile is dirty?

Gotcha, I can replace the image with a label. Highlighting the button sounds interesting too, I'll add it to the list.

whot:
If you can't do that in the widgets, you could do this in the dbus bindings. Set a dirty flag on the device whenever you change a property, you can than query it.  That's the quick and dirty approach :)

> what happens if the user wants to abort the changes they made?
That's what the cancel buttons are for, once you click "OK" on a button, there's no abort anymore. If we remove the Save button altogether, then the expectation of being able to abort at any time goes away.

bentiss:
Also, automated commit is basically what happens under the other operating system with the proprietary applications.

The only situation where it could be an issue is if you remap buttons and forget to have a primary click somewhere. This can be detected and we can then either automatically switch the other primary physical button (think left/right button type) to primary logical button action. We could also show a popup warning the user.

Hjdskes:
[...]

PR #101, opened 2017-08-10
Make the commit button more obvious

Hjdskes:
This PR implements some of @hadess' suggestions from #69. I think that timeout-based commit is something we should look into and is perhaps something for a future version of Piper, but for now as a stop-gap sort of thing I went with these suggestions:

1. Disable the button when nothing has been changed
2. Make the button blue like the default button in dialogues ('suggested-action' style IIRC) when something has been changed

As for the icon, personally I'm not confused by it but if you feel that others might I'm open to switching it for something else. Perhaps we can make a custom save icon, that replaces the disk with a mouse? In any case, I much prefer an icon over text so I'd rather do that as a last resort.

The other suggestions about asking users about unsaved changes before exiting are not related to the button per sé so I'll open separate issues for those, because I do like them and want to implement them.

As for activity indicators, I think that all our operations are fast enough not to require those, but that might be my device only? In any case, this would also be something for the future. Feel free to open an issue for this yourself.

@hadess if you're reading this, I'd appreciate your comments on these changes!

### DIFF
--- a/data/ui/MousePerspective.ui
+++ b/data/ui/MousePerspective.ui
@@ -120,7 +120,7 @@
       </object>
     </child>
     <child>
-      <object class="GtkButton">
+      <object class="GtkButton" id="button_commit">
         <property name="visible">True</property>
         <property name="can_focus">True</property>
         <property name="receives_default">True</property>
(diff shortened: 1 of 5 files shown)


whot:
wfm, we can change this again if need be

hadess:
Commented on the icon in the related, still opened, issue.

PR #103, opened 2017-08-11
MousePerspective: change button icon to text

### LINKED ISSUE #69: Save button is not obvious

Hjdskes:
Fixes #69.

@hadess [is right](https://github.com/libratbag/piper/issues/69#issuecomment-321776046).


hadess:
Commit is either an overly technical term, or a promise to do something. Any reason not to use "Apply"?
I would not recommend using "Save" as it's likely to be misconstrued as saving to the machine's internal storage, not applying the settings to the mouse.

Hjdskes:
Fixed by amending

PR #104, opened 2017-08-11
Ask for confirmation when quitting with unsaved changes

### LINKED ISSUE #102: Confirm before exiting with uncommitted changes
As pointed out by @hadess in #69:

* Ask the user whether to apply all the settings to the mouse before exiting (same pattern as when you exit gedit with an unsaved file)
* Add a preference to automatically save to the mouse when exiting (default to yes?)

I definitely want to implement the first; I'm not sure if the second is worth it.

Hjdskes:
This adds another property to the perspectives, which signals whether a perspective can safely shutdown. On a delete event, the Window checks all its perspectives' properties and if one perspective signals that it cannot safely shutdown, a dialog is presented that allows the user to either shutdown ignoring any changes or to abort and commit or make other changes.

Fixes #102.

### DIFF
--- a/piper/window.py
+++ b/piper/window.py
@@ -23,7 +23,7 @@ from .welcomeperspective import WelcomePerspective
 
 import gi
 gi.require_version("Gtk", "3.0")
-from gi.repository import GLib, Gtk
+from gi.repository import Gdk, GLib, Gtk
 
 
 @GtkTemplate(ui="/org/freedesktop/Piper/ui/Window.ui")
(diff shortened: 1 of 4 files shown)


piper/window.py
```diff
@@ -62,6 +62,20 @@ def __init__(self, ratbag, *args, **kwargs):
         else:
             self._present_welcome_perspective(ratbag.devices)
 
+    def do_delete_event(self, event):
+        for perspective in self.stack_perspectives.get_children():
+            if not perspective.can_shutdown:
+                dialog = Gtk.MessageDialog(self, Gtk.DialogFlags.MODAL,
+                                           Gtk.MessageType.QUESTION,
+                                           Gtk.ButtonsType.YES_NO,
+                                           _("There are unsaved changes. Are you sure you want to quit?"))
```
whot: if we use "apply" for the button we should use "unapplied" here too instead of unsaved.

whot:
please rebase, fix the minor issue and merge, thanks

Issue #82, opened 2017-08-01
Don't allow unsetting the left button


whot:
If we map current logical left button to anything else, map another button to left instead. Default approach: search for the right button and swap it. Otherwise pick some random button (?).

We don't want to end up without a left button

Hjdskes:
Isn't this something we should do in ratbagd/libratbag?

whot:
I don't think so. ratbagd is a configuration proxy but shouldn't really enforce behaviour. for piper it makes sense to disallow users to lock themselves out, but that's not something to be enforced at the lower levels (also, we don't have the same feeback/error reporting/etc. in ratbagd)

Hjdskes:
I see, those are good points. In that case this is something that I'd rather to in Piper's ratbagd bindings than the actual GUI code itself (e.g. in the ButtonsPage). That might mark the point where ratbatctl's and Piper's bindings diverge, though. What do you think?

whot:
hmm, depends on your implementation. i can't quite picture how doing that in the bindings is going to be the best solution, it seems the wrong place for it. This seems like something that should be handled in the button assignment box, because that's the only place it matters

Hjdskes:
>  This seems like something that should be handled in the button assignment box, because that's the only place it matters

I suppose it depends on the solution that we apply here. The one below naturally belongs in the GUI :)

So now the question is, what do we do when the user remaps the left mouse button? Obviously there are perfectly valid scenarios to do so (e.g. switching left- and right mouse buttons for left-handed people who prefer to control the device with their left hand (I'm left handed but mouse with my right hand)), so we shouldn't outright disallow it but mapping the right mouse button to a left click by default also feels sketchy. I'm leaning towards popping up a dialog that asks the user if they know what they're doing *when we detect there isn't another button mapped to a left mouse click when they are attempting to remap the only/last button that is mapped to a left mouse click*, something like:

```
+---------------------------------------------------------------------+
| You are currently attempting to remap a button that will leave you  |
|         without a left mouse click. Do you want to continue?        |
|                                                                     |
|                                                            Yes / No |
+---------------------------------------------------------------------+
```

This shouldn't be difficult and be only a few additional lines of code, while I think it's the least intrusive solution. Ok, it is another dialog that we should generally avoid but 1) we cannot clean this up after the fact so an "undo" action is not an option, and 2) we avoid undiscoverable and intrusive behavior that would occur if we remap a random button to the left mouse click.

Hjdskes:
[...]

PR #117, opened 2017-08-15
ButtonDialog: only allow swapping left- and right buttons

Hjdskes:
Another take for #82; where we only allow swapping the left- and right mouse button. Until https://github.com/libratbag/libratbag/issues/233 is fixed I'm using the buttons' indices to detect whether a button is left or right (with approval from @whot).

### DIFF
--- a/data/ui/ButtonDialog.ui
+++ b/data/ui/ButtonDialog.ui
@@ -69,7 +69,7 @@
                                 <property name="width_request">100</property>
                                 <property name="visible">True</property>
                                 <property name="can_focus">True</property>
-                                <property name="tooltip_text" translatable="yes">Add a new resolution to the profile</property>
+                                <property name="tooltip_text" translatable="yes">Capture a macro for this button</property>
                                 <child>
                                   <object class="GtkBox">
                                     <property name="visible">True</property>
--- a/data/ui/ButtonDialog.ui
+++ b/data/ui/ButtonDialog.ui
@@ -106,6 +106,7 @@
                                         <property name="can_focus">False</property>
                                         <property name="halign">start</property>
                                         <property name="label" translatable="yes">Send keystroke</property>
+                                        <property name="track_visited_links">False</property>
                                       </object>
                                       <packing>
                                         <property name="expand">False</property>
--- a/data/ui/ButtonDialog.ui
+++ b/data/ui/ButtonDialog.ui
@@ -47,7 +47,7 @@
                 <property name="orientation">vertical</property>
                 <property name="spacing">6</property>
                 <child>
-                  <object class="GtkScrolledWindow">
+                  <object class="GtkScrolledWindow" id="button_mappings">
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="shadow_type">in</property>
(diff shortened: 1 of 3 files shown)


piper/buttondialog.py
```diff
@@ -226,6 +240,9 @@ def _on_row_activated(self, listbox, row):
                 print("Unable to grab keyboard, can't set keystroke", file=sys.stderr)
             else:
                 self.stack.set_visible_child_name("capture")
+        elif row is self.row_left_hand_mode:
+            self._action_type = RatbagdButton.ACTION_TYPE_BUTTON
+            self._mapping = -2
```
Hjdskes: @whot (or any other reviewer): this part in particular I am unhappy with. There needs to be some kind of value to signal to the `ButtonsPage` that we want to swap buttons, and so for now I've decided on -2 because I can envision -1 being used by libratbag at some point, but every value is quite arbitrary. I'm not sure if there is any other, better approach that we can use without having to redo this whole dialog.

data/ui/ButtonDialog.ui
```diff
@@ -106,6 +106,7 @@
                                         <property name="can_focus">False</property>
                                         <property name="halign">start</property>
                                         <property name="label" translatable="yes">Send keystroke</property>
+                                        <property name="track_visited_links">False</property>
```
whot: is this just detritus or intentional?

data/ui/ButtonDialog.ui
```diff
@@ -250,4 +251,31 @@
...
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="margin_top">4</property>
+        <property name="margin_bottom">4</property>
+        <child>
+          <object class="GtkLabel">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">start</property>
+            <property name="label" translatable="yes">Switch the left- and right mouse buttons</property>
```
whot: I think this should be "Swap the left and right mouse buttons", see https://english.stackexchange.com/questions/22954/is-there-any-difference-between-switch-and-swap for a (non-authoritative) explanation between switch and swap.

But what I'm not happy with here is that if I click on the mouse button, the only option I have now is to swap and my only recourse is cancel. This is a bit confusing, shouldn't there be an entry for the current option too? i.e. right-handed?

I'm not happy with it anyway, it's not expressive enough, but I don't know how to make it better with this design. I'm starting to think it may be worth having a different dialog style here and for the right button that only offers left/right click and possibly a little graphic or explanation. Basically what the Mouse&Touchpad panel in the control center does.

piper/buttondialog.py
```diff
@@ -96,13 +97,26 @@ def _init_ui(self, buttons):
...
-        for button in buttons:
-            key, name = self._get_button_name_and_description(button)
-            row = ButtonRow(name, _("Button mapping"), RatbagdButton.ACTION_TYPE_BUTTON, button.index + 1)
-            self.listbox.insert(row, i)
-            if self._action_type == RatbagdButton.ACTION_TYPE_BUTTON and button.index + 1 == self._button.mapping:
-                self.listbox.select_row(row)
+        # TODO: make this self._button.type == RatbagdButton.TYPE_LEFT or
+        # or self._button.type == RatbagdButton.TYPE_RIGHT once
+        # https://github.com/libratbag/libratbag/issues/233 is fixed.
+        if self._button.index == 0 or self._button.index == 1:
```
whot: note to self, mostly: I wonder how ```if self._button.index in [0, 1]``` would look like here :)

piper/buttondialog.py
whot: you should make it a const value nonetheless (like a C #define), that way you only have one thing to change in the future. and it's less magic than a -2

Generally I recommend that when you're re-using something that's otherwise a fixed range, choose something well outside that range, e.g. -2000. still obviously visible as non-garbage but much easier to grep for than -2 and less likely to get used by the original value. that's also why the ratbag error codes start at 1000 (errno overlap)

piper/buttondialog.py
```diff
@@ -96,13 +97,26 @@ def _init_ui(self, buttons):
...
+            self.row_left_hand_mode._section = _("Button mapping")
+            self.listbox.insert(self.row_left_hand_mode, i)
+            if self._action_type == RatbagdButton.ACTION_TYPE_BUTTON and \
+                    (self._button.index + 2 == self._button.mapping or
+                        self._button.index == self._button.mapping):
```
whot: I admit, it's not obvious where the +2 comes from, sorry. What am I missing here?

data/ui/ButtonDialog.ui
Hjdskes: Intentional, but I'll split it into its own commit. I missed it in the diff.

data/ui/ButtonDialog.ui
Hjdskes: > I think this should be "Swap the left and right mouse buttons", see https://english.stackexchange.com/questions/22954/is-there-any-difference-between-switch-and-swap for a (non-authoritative) explanation between switch and swap.

Thanks, in hindsight that's obvious :)

> But what I'm not happy with here is that if I click on the mouse button, the only option I have now is to swap and my only recourse is cancel. This is a bit confusing, shouldn't there be an entry for the current option too? i.e. right-handed?

I see what you mean, but the option "toggles" basically. If you currently have the mouse in right-hand mode, it switches to left-hand mode and vice versa. I agree though that it's not ideal.

> I'm starting to think it may be worth having a different dialog style here and for the right button that only offers left/right click and possibly a little graphic or explanation. Basically what the Mouse&Touchpad panel in the control center does.

Hmm... I can see where you're going with this. To be clear: you want separate list boxes for each type of mapping? That could work out nicely, as I can then just replace the button mapping listbox with a widget + possible graphical or textual explanation. I'll whip up a proof of concept today.

piper/buttondialog.py
Hjdskes: It'd look nice, hint taken ;)

piper/buttondialog.py
Hjdskes: I wanted to add a comment here but decided against it. Wrong choice, hehe.

This `if` is there to highlight the current row if this option is currently set on the device. Since we're using the button's index as a replacement for its type (https://github.com/libratbag/libratbag/issues/233), we need to compare the index to the actual button mapping to see if it's a left/right mouse button assigned to left/right mouse click. Two scenarios:

1. Left mouse button (index 0) is mapped to right mouse button, where mappings are 1-indexed and thus right mouse click has value 2: `self._button.index + 2 == self._button.mapping`.
2. Right mouse button (index 1) is mapped to left mouse button, where mappings are again 1-indexed and thus left mouse click has value 1: `self._button.index == self._button.mapping`.

If this code sticks around with the new dialog I'll add a comment :)

data/ui/ButtonDialog.ui
whot: I think I want two dialogs, the current one for all buttons but left/right and one that essentially just has a on/off with an explanation below/next to it/wherever. So what you propose, swapping out the listbox with the other widget is the right thing, yes.

data/ui/ButtonDialog.ui
Hjdskes: ^ Done.

piper/buttondialog.py
```diff
@@ -230,6 +258,16 @@ def _on_row_activated(self, listbox, row):
             self._action_type = row._action_type
             self._mapping = row._value
```
whot: I'm gonna have to be annoying here but: this is not good style. If you have a function that does one of two things based on a condition and **both** blocks are reasonably complicated (i.e. more than two lines), split it into helper functions and call them like this:
````
def init_ui():
   if button.index in [0, 1]:
     self._init_primary_buttons_ui()
   else:
     self._init_other_buttons_ui()
````
Much more readable than the semi-hidden `return` you have here.

Hjdskes: Thanks, I should've known.

Issue #120, opened 2017-08-15
Piper crashes when opening devices with weird profiles.


cvuchener:
I have some mice where I tested weird profiles. Piper crashes when I try to open them.

First one is a G500 using "Switch to profile X" special action that is not supported by libratbag ("UNKNOWN").

```
Traceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/gi/_propertyhelper.py", line 427, in obj_set_property
    prop.fset(self, value)
  File "/usr/local/lib/python3.6/site-packages/piper/mouseperspective.py", line 89, in device
    self.stack.add_titled(ButtonsPage(device), "buttons", _("Buttons"))
  File "/usr/local/lib/python3.6/site-packages/piper/buttonspage.py", line 50, in __init__
    self._set_profile(self._device.active_profile)
  File "/usr/local/lib/python3.6/site-packages/piper/buttonspage.py", line 57, in _set_profile
    self._on_ratbagd_button_action_type_changed(ratbagd_button, None, button)
  File "/usr/local/lib/python3.6/site-packages/piper/buttonspage.py", line 83, in _on_ratbagd_button_action_type_changed
    label = RatbagdButton.SPECIAL_DESCRIPTION[ratbagd_button.special]
KeyError: 4294967295
```

The second is a G500s with multiple profiles (but libratbag only supports one profile for the G500s). I am not sure what is the exact cause for the crash.

```
Traceback (most recent call last):
  File "/usr/lib64/python3.6/site-packages/gi/_propertyhelper.py", line 427, in obj_set_property
    prop.fset(self, value)
  File "/usr/local/lib/python3.6/site-packages/piper/mouseperspective.py", line 87, in device
    self.stack.add_titled(ResolutionsPage(device), "resolutions", _("Resolutions"))
  File "/usr/local/lib/python3.6/site-packages/piper/resolutionspage.py", line 66, in __init__
    self._init_ui()
  File "/usr/local/lib/python3.6/site-packages/piper/resolutionspage.py", line 86, in _init_ui
    self._on_active_profile_changed(self._device, profile)
  File "/usr/local/lib/python3.6/site-packages/piper/resolutionspage.py", line 92, in _on_active_profile_changed
    self.rate_500.set_active(active_resolution.report_rate == 500)
AttributeError: 'NoneType' object has no attribute 'report_rate'
```

Output of `ratbag-command info`:

```
ratbag info: Profile 0 has a wrong CRC, assuming valid.
[...]

PR #122, opened 2017-08-15
Do not crash if there is no active profile or resolution

Hjdskes:
See #120.

I'm not entirely sure about this, because basically we're working around broken device drivers here. If the driver is broken, it doesn't make much sense to try and continue. Maybe instead I should add a check for an active profile and resolution when the MousePerspective is being opened, and bail from there only. Or, we let things as-is and make sure our drivers work.


Hjdskes:
> just hide this in the dbus bindings. this is clearly a bug in ratbagd so make the bindings return the first profile/resolution on failure together with the printf and leave the rest of piper blissfully ignorant :)

D'oh, that's a much better solution, thanks.

PR #131, opened 2017-08-16
Meson: simplify development version

Hjdskes:
Now we only have one configuration structure, and only one file will be generated instead of two.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,6 +34,7 @@ config_piper.set('pkgdatadir', pkgdatadir)
 config_piper.set('localedir', localedir)
 config_piper.set('gtk_major_version', gtk_major_version)
 config_piper.set('gtk_minor_version', gtk_minor_version)
+config_piper.set('devel', '')
 
 configure_file(input: 'piper.in',
 	       output: 'piper',
--- a/meson.build
+++ b/meson.build
@@ -39,7 +39,7 @@ config_piper.set('devel', '')
 configure_file(input: 'piper.in',
 	       output: 'piper',
 	       configuration: config_piper,
-	       install_dir: bindir )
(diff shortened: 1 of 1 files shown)


Hjdskes:
@whot if you don't like the new meson.build, at least cherry-pick 91e3fcbc213928940c4b6caab06042cf3786b067.

meson.build
```diff
@@ -34,11 +34,12 @@ config_piper.set('pkgdatadir', pkgdatadir)
 config_piper.set('localedir', localedir)
 config_piper.set('gtk_major_version', gtk_major_version)
 config_piper.set('gtk_minor_version', gtk_minor_version)
+config_piper.set('devel', '')
```
whot: I can imagine what this does, but please add the failure message to the commit message, it makes it much easier to a) understand what's going on and b) identify if any other future failures are caused by separate issues or by an insufficient patch.

meson.build
```diff
@@ -34,11 +34,12 @@ config_piper.set('pkgdatadir', pkgdatadir)
...
 
 configure_file(input: 'piper.in',
 	       output: 'piper',
 	       configuration: config_piper,
-	       install_dir: bindir )
+	       install_dir: bindir)
```
whot: for whitespace changes: please keep them in separate patches and push them directly.

whot:
I filed https://github.com/mesonbuild/meson/issues/2206 to make this easier in meson

Hjdskes:
> I filed https://github.com/mesonbuild/meson/issues/2206 to make this easier in meson

Thanks, that should also make things simpler. Let's go with that if/when that gets merged. I've rebased this branch to only contain the two commits you wanted; will merge that when the check is done.

Issue #81, opened 2017-08-01
Switch active resolution on resolution adjustments


whot:
When we click on a resolution and start changing it, we should switch to that resolution so that the user sees the effects once the auto-apply runs in the background.

Hjdskes:
#132 is now in but the resolution doesn't actually apply before the "Apply" button is clicked. That's just the way it is right now. Feel free to reopen if I should actually change this and commit behind the user's back.

Issue #72, opened 2017-07-30
changing a resolution should not change the active one


bentiss:
When testing a device that has resolution 2 active, changing any other resolution in the Resolution panel and committing to the device changes the active resolution back to the first one (0).

I do not have this behavior with ratbagctl so I guess piper must be overwriting the active resolution...

Hjdskes:
I cannot reproduce this here. In any case, no code related to resolutions (`resolutionspage.py`, `resolutionrow.py`) calls `RatbagdResolution::set_active` at any time. 

Are you sure you weren't doing something else?

Hjdskes:
Also this seems to conflict with https://github.com/libratbag/piper/issues/81

whot:
I'd prefer the #81 behaviour, it provides immediate feedback to what the user just set. @bentiss any comments?

Hjdskes:
Closed because #132 is merged.

PR #132, opened 2017-08-16
ResolutionRow: add label to highlight the active resolution

Hjdskes:
The same approach should work for showing and setting the default resolution (see #36) so I'll look into fixing that too. Note that this depends on an incoming patch to ratbagd.

### DIFF
--- a/data/ui/ResolutionRow.ui
+++ b/data/ui/ResolutionRow.ui
@@ -58,6 +58,21 @@
                 <property name="position">1</property>
               </packing>
             </child>
+            <child>
+              <object class="GtkLabel" id="active_label">
+                <property name="can_focus">False</property>
+                <property name="halign">end</property>
+                <property name="label" translatable="yes">Active</property>
+                <property name="justify">right</property>
+                <property name="track_visited_links">False</property>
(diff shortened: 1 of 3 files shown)


Hjdskes:
See https://github.com/libratbag/libratbag/pull/284.

Hjdskes:
Whoops, I forgot to mention that this also fixes #81 and therefore overrides #72, although I still don't get how that can work given that the changes do have to be committed first.

whot:
feel free to push the first three commits, they're semi-unrelated and can go in

Hjdskes:
> feel free to push the first three commits, they're semi-unrelated and can go in

Thanks, done. Looking into the bug now.

Hjdskes:
I cannot reproduce your bug @whot. I have two resolutions in a single profile both with a DPI of 2700, one of which is active. If I start Piper, only the active one is labelled active -- the other isn't. Are you sure this isn't something on your end, where ratbagd/libratbag/your device has two resolutions active?

whot:
merged, will file an issue if it's still broken here but meanwhile let's get this in

PR #136, opened 2017-08-17
Introduce RatbagdMacro to abstract macros, and fix their display in the ButtonsPage

Hjdskes:
This adds a property that translates the macro from keycodes to a human readable
string representation, so that it can be used to present the macros to
the user in the `ButtonsPage`, and applies that property in the `ButtonsPage` as well.

It does introduce some duplication between `RatbagdButton::macro_str` and `KeyStroke::_update_macro`. We can fix this by making `macro_str` a static method instead, but I wasn't sure if that was worth it.


Hjdskes:
Just FYI and as a reminder to myself, the plan here is to add a `RatbagdMacro` class to the bindings that will take care of the string and yada yada; currently trying to see if I can move `KeyStroke` into `RatbagdMacro`.

piper/keystroke.py
```diff
@@ -111,11 +111,12 @@ def _update_macro(self):
...
-            keyval = ecodes.KEY[keycode]
+        for (type, val) in self._keys:
             if type == RatbagdButton.MACRO_KEY_PRESS:
-                keys.append("+{}".format(keyval))
+                keys.append("↓{}".format(ecodes.KEY[val]))
             elif type == RatbagdButton.MACRO_KEY_RELEASE:
-                keys.append("-{}".format(keyval))
+                keys.append("↑{}".format(ecodes.KEY[val]))
+            elif type == RatbagdButton.MACRO_WAIT:
+                keys.append("{:.2f}s".format(val / 1000.0))
```
Hjdskes: @whot wanted this in seconds, not milliseconds (reminder to self).

Hjdskes:
@whot This is ready for review again

piper/ratbagd.py
```diff
@@ -658,23 +660,19 @@ def mapping(self, button):
...
+        object.
 
-        @param macro A list of (type, value) tuples to form the new macro.
+        @param macro A RatbagdMacro object representing the macro to apply to
+                     the button, as RatbagdMacro.
         """
-        self._dbus_call("SetMacro", "a(uu)", macro)
-        self._set_dbus_property("Macro", "a(uu)", macro, readwrite=False)
+        self._dbus_call("SetMacro", "a(uu)", macro.keys)
+        self._set_dbus_property("Macro", "a(uu)", macro.keys, readwrite=False)
```
Hjdskes: Yea this had yet to be rebased on top of that rework.

piper/ratbagd.py
```diff
@@ -790,20 +789,23 @@ def cancel(self):
...
+                print("Keycode is not within the valid range.", file=sys.stderr)
+                return
+            value = value - self._XORG_KEYCODE_OFFSET
         # Only append if the entry isn't identical to the last one, as we cannot
         # e.g. have two identical key presses in a row.
-        if len(self._macro) == 0 or (type, keycode) != self._macro[-1]:
-            self._macro.append((type, keycode))
+        if len(self._macro) == 0 or (type, value) != self._macro[-1]:
+            self._macro.append((type, value))
             self.notify("keys")
```
Hjdskes: This commit does actually add a line (line 800); the renaming is done to indicate that we treat more than just keycodes now. As such I'd prefer to keep it separate, but it's your call :)

piper/ratbagd.py
```diff
@@ -725,6 +723,90 @@ def disable(self):
...
+    def append(self, type, keycode):
+        """Appends the given event to the current macro.
+
+        @param type The type of event, as one of RatbagdButton.MACRO_*.
+        @param keycode The keycode of the event, as int in X.Org or Gdk value.
+        """
+        if not self._XORG_KEYCODE_OFFSET <= value <= 255:
+            print("Keycode is not within the valid range.", file=sys.stderr)
+            return
+        keycode = keycode - self._XORG_KEYCODE_OFFSET
```
Hjdskes: Woops, this should've been put in ButtonDialog indeed. Thanks for catching it.

piper/ratbagd.py
Hjdskes: After moving the keycode handling up to the caller this commit indeed doesn't make sense anymore.

PR #137, opened 2017-08-17
Add localization support

Hjdskes:


Hjdskes:
I went ahead and added the internationalization introduced by @whot in https://github.com/whot/piper/commit/44d8265d4a973ec45962d75f79e98fab1925b6ff, minus my comments. The only issue seems to be that the translations from the maps in the bindings aren't being applied, I'm yet to figure out why.

Hjdskes:
It's currently not possible to run `./build/piper.devel` on local translations to for example test a translation, because Meson places the `$lang.gmo` files in `build/po/$lang.gmo` and this isn't what gettext expects. It works if you make it e.g. `build/po/nl_NL/LC_MESSAGES/piper.mo` and run e.g. `LC_ALL=nl_NL.UTF-8 ./build/piper.devel`, but I don't think there is a nice way of making Meson do this.

So, for now at least, to test translations you need to actually install Piper (`sudo ninja -C build install` and go from there, with e.g. `LC_ALL=nl_NL.UTF-8 piper`.

circle.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       # now install and checks the resulting file (just in case)
       - run:
           command: |
-            dnf install -y meson python3-devel pygobject3-devel
+            dnf install -y meson gettext python3-devel pygobject3-devel
```
whot: this patch needs to be rebased to before when we need it (or just squashed in with the first commit that needs it)

po/de_AT.po
```diff
@@ -0,0 +1,546 @@
+# German Piper translation
+# Copyright (C) 2017 Red Hat, Inc
+# This file is distributed under the same license as the piper package.
+# Peter Hutterer <peter.hutterer@who-t.net>, 2017.
+#
+msgid ""
```
whot: this file seems lost here. let's bring it back to the right commit

Hjdskes: Ehm, not sure how this got added. I'll remove it :)

whot:
regarding the piper.devel issue: I'm willing to make this meson's problem, this doesn't sound like something that's just our problem. Especially since ninja build doesn't actually build the .gmo files, only ninja install does it as part of the gettext invocations.

But at the same time, I don't consider it a huge issue, live-testing of translations isn't quite as urgent, so this shouldn't hold things back.

Hjdskes:
Before we merge this I'd also like to add more context for translators for certain strings. @whot you are likely a good candidate to test which strings (definitely) need context; I know which strings are used where...

po/de_AT.po
Hjdskes: I'll update it and add it in its own commit.

Hjdskes:
Also, finally, we should merge this once all UI work is done so we only translate everything once.

circle.yml
Hjdskes: Done, even though for some reason it's listed in the wrong order here. [It shows correct in my branch](https://github.com/Hjdskes/piper/commits/fix/localization).

piotrdrag:
The “_NL” part in po/nl_NL.po is redundant. Just “po/nl.po” is preferable in this case.

Hjdskes:
> The “_NL” part in po/nl_NL.po is redundant. Just “po/nl.po” is preferable in this case.

Thanks! 

I see you are contributing translations to other projects, so if you don't mind me asking: is there the general preference among translators to do their work? As you may have seen in this PR, right now we rely on translators generating the .pot and .po files manually; we are aware there are tools to do all this work for them, but since we don't have any experience translating we figured we'd start like this and adopt whatever the (majority of) translators (if any, we're being hopeful here :D) prefer to use.

Also, have you perhaps seen the issue of translations not being applied from Python maps before? For example, in a Python map like the following the non-translated English string `One` is used instead of the translation:

```python
example_map = {
  1: _("One"),
}
```

You can clearly see this image in [this video I recorded for my blog](https://hjdskes.github.io/img/blog/gsoc-part-13/piper-dutch.webm) where the English text is used even though all of those strings are translated.

piotrdrag:
>I see you are contributing translations to other projects, so if you don't mind me asking: is there the general preference among translators to do their work? As you may have seen in this PR, right now we rely on translators generating the .pot and .po files manually; we are aware there are tools to do all this work for them, but since we don't have any experience translating we figured we'd start like this and adopt whatever the (majority of) translators (if any, we're being hopeful here :D) prefer to use.

What you’re doing right now is definitely too high of an entry barrier. I wouldn’t expect many translators to be able to work this way.

My personal preference would be to move to the GNOME infrastructure and [Damned Lies](https://l10n.gnome.org/), and get dozens of translations for free from our great translation teams.

There are other ways, of course. [Transifex](https://www.transifex.com/) is popular among projects on GitHub, but you’d need to build your translation teams from scratch.

>Also, have you perhaps seen the issue of translations not being applied from Python maps before? For example, in a Python map like the following the non-translated English string One is used instead of the translation:

>example_map = {
  1: _("One"),
}

>You can clearly see this in this video I recorded for my blog where the English text is used even though all of those strings are translated.

I can’t be sure, as I don’t code, but I think you may want to use the N_ macro for these.

https://developer.gnome.org/glib/stable/glib-I18N.html#N-:CAPS

https://docs.python.org/3/library/gettext.html#deferred-translations

Hjdskes:
> My personal preference would be to move to the GNOME infrastructure and Damned Lies, and get dozens of translations for free from our great translation teams.

Is there some information on what it would take to get this to work? Maybe infrastructure stuff is something @whot should handle though.

> I can’t be sure, as I don’t code, but I think you may want to use the N_ macro for these.

You're spot on, thank you very much! :)

piotrdrag:
>>My personal preference would be to move to the GNOME infrastructure and Damned Lies, and get dozens of translations for free from our great translation teams.

>Is there some information on what it would take to get this to work? Maybe infrastructure stuff is something @whot should handle though.

Here’s some starting information: https://wiki.gnome.org/Git/NewRepository

whot:
> Here’s some starting information: https://wiki.gnome.org/Git/NewRepository

AFAICT those are the instructions to move piper to the gnome infrastructure altogether, away from github. I don't think that's a particularly good idea right now. Long-term maybe, but definitely not while we're in the middle [1] of GSoC. And we still have a number of libratbag issues that require some tight coupling between libratbag and piper - having both on github as part of the same project helps here.

Furthermore, from some more traversing of the gnome wiki [2], the actual translation is still done on .pot files, right? This isn't any different to just doing this here. The biggest barrier to entry right now would be building the initial .pot file, any file that is translated is checked in and can be edited directly. And the pot file is too high a barrier, we can add that to git.


[1] well, more like at the end of...
[2] https://wiki.gnome.org/TranslationProject/LocalisationGuide

po/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+This directory contains the translations of Piper.
+
+For errors in translations, please [file an
+issue](https://github.com/libratbag/piper/issues/new).
+
+New or updated translations are always welcome. To start a new translation, run:
+
+	$ ninja -C $builddir piper-pot
+	$ cp po/piper.pot po/$lang.po
```
whot: two things here: make it with a meson command so you can use less variables like `$builddir`. And the `cp` won't work because it's in the builddir. Make it something like:

````
  $ meson translation-build
  $ ninja -C translation-build piper-pot
  $ cp translation-build/po/piper.pot po/$lang.po
  # you can remove the build directory now
  $ rm -rf translation-build
  # add the new file to git
  $ git add po/$lang.po
  # now edit the file and add translations
````

File: circle.yml (hunk 1 above)
Comment (whot) [inline circle.yml]: just squash this in with the Add localization support commit please, I think that's easier than a standalone commit that doesn't explain much. It's  just a single line here anyway

File: data/ui/piper.ui [hunk 4]
```diff
@@ -1,927 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.19.0 -->
```
whot: I think it's best if you extract that commit to master, git push it and then rebase the translation branch on top. otherwise this is mixed in with the translation bits, which is a bit odd.

whot:
I forgot to mention in the review commit: feel free to push the commits up to the circleci commit (i.e. the UI changes and the piper removal commit) to master and only leave the rest in this branch

whot:
> It's currently not possible to run ./build/piper.devel on local translations to for example test a translation, because Meson places the $lang.gmo files in build/po/$lang.gmo and this isn't what gettext expects. 

ftr: https://github.com/mesonbuild/meson/pull/2227

po/README.md
whot: doh, just noticed that piper.pot is generated in the source dir. how strange...

circle.yml
Hjdskes: Done.

Hjdskes:
> I forgot to mention in the review commit: feel free to push the commits up to the circleci commit (i.e. the UI changes and the piper removal commit) to master and only leave the rest in this branch

Done, thanks.

Hjdskes:
@whot I just rebased this on top of #150 so you don't have to fix any conflicts. I added some context where I think it's needed; it would be nice if you could update the German translation you did and see if there's any context missing.

piotrdrag:
>> Here’s some starting information: https://wiki.gnome.org/Git/NewRepository

>AFAICT those are the instructions to move piper to the gnome infrastructure altogether, away from github. I don't think that's a particularly good idea right now. Long-term maybe, but definitely not while we're in the middle [1] of GSoC. And we still have a number of libratbag issues that require some tight coupling between libratbag and piper - having both on github as part of the same project helps here.

I understand. I’m afraid, however, that https://l10n.gnome.org/ works only with https://git.gnome.org/.

I also believe that such a nice GTK+ app as Piper would feel right at home on GNOME infra when the time comes. :)

>Furthermore, from some more traversing of the gnome wiki [2], the actual translation is still done on .pot files, right? This isn't any different to just doing this here. The biggest barrier to entry right now would be building the initial .pot file, any file that is translated is checked in and can be edited directly. And the pot file is too high a barrier, we can add that to git.

Yes, it’s .pot/.po files, but Damned Lies makes downloading and uploading translations easier for non-technical translators (which is most of us) by an order of magnitude.

whot:
rebased locally and merged as part of eed44e594222d

whot:
> Yes, it’s .pot/.po files, but Damned Lies makes downloading and uploading translations easier for non-technical translators (which is most of us) by an order of magnitude.

understood. piper can move to gnome at some point in the future, but right now it's still a bit of a wild ride :)

PR #143, opened 2017-08-18
ButtonDialog: add a search entry to search mappings

Hjdskes:
This required a bit of changes to the ButtonDialog's UI, so I decided to overhaul it a bit more and hopefully improve its looks.

### DIFF
--- a/data/ui/ButtonDialog.ui
+++ b/data/ui/ButtonDialog.ui
@@ -40,242 +40,96 @@
             <property name="visible">True</property>
             <property name="can_focus">False</property>
             <child>
-              <object class="GtkBox">
+              <object class="GtkScrolledWindow">
                 <property name="visible">True</property>
-                <property name="can_focus">False</property>
-                <property name="orientation">vertical</property>
+                <property name="can_focus">True</property>
+                <property name="shadow_type">in</property>
+                <property name="min_content_height">250</property>
+                <property name="max_content_width">0</property>
+                <property name="max_content_height">250</property>
+                <property name="propagate_natural_width">True</property>
+                <property name="propagate_natural_height">True</property>
                 <child>
-                  <object class="GtkScrolledWindow" id="button_mappings">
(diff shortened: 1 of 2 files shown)


data/ui/ButtonRow.ui
```diff
@@ -9,10 +9,10 @@
       <object class="GtkBox">
         <property name="visible">True</property>
         <property name="can_focus">False</property>
-        <property name="margin_left">6</property>
-        <property name="margin_right">6</property>
-        <property name="margin_top">4</property>
-        <property name="margin_bottom">4</property>
```
whot: I'm not sure this is right, tbh. It should be the same layout as the list in the Keyboard control-center panel and it seems just those few pixels larger now (hard to tell though). what's the specific reason for this?

Hjdskes: The thing is that they don't look the same with these values, which is why I tried to increase them. If you compare the original to the control-center's keyboard panel, this is what you see:

screenshot from 2017-08-21 11-18-05

versus

screenshot from 2017-08-21 11-18-19

Compare that to the new values:

screenshot from 2017-08-21 11-20-15

According to GtkInspector, the new values give an allocation of 35 pixels in height versus 38 in the keyboard panel so it's still not exactly the same. I suppose their height is larger due to the image.

Hjdskes:
> I'd say let's skip the commit that moves things around

You mean 88d009fe12c80d70eb977ec7e8dd4fb9c3d13ce9? That doesn't just move things around but also reworks the widget. Have you seen that part of it too?

whot:
sorry, I meant 4150096a76574eb16213fe8854782d949c97da49 (it moves around the margins but nothing else, so it's easy to skip). As a general rule: designers are more antsy about margins that we will ever be, and they know the details of which ones to use where. So I find it best to not worry too much and wait for someone to give you exact numbers :)

PR #150, opened 2017-08-21
Limit length of displayed macros

### LINKED ISSUE #144: Macros with successive button press and release shouldn't display both
Rather, we should simply display the key once with ↕, because without this the list can get *very* long quite easily.

Hjdskes:
Fixes #144, should go after #136.

I'm not particularly happy with the `__str__` method in `RatbagdMacro`, but it's been a long day. Maybe I can improve it tomorrow.


whot:
This one was merged as part of 98a14a9e64bf4f5a8a1aa524e5dfc55922609504, I just pushed it directly from your branch.

PR #155, opened 2017-08-23
Add Polish translation

piotrdrag:

PR #156, opened 2017-08-23
Some more context and comments for translators

Hjdskes:
From #155 (see the comments there).


Hjdskes:
@piotrdrag can you check the two fuzzy strings in the Polish translation? I did my best to fix the pl.po file but I don't speak Polish..

piper/buttondialog.py
```diff
@@ -220,6 +220,7 @@ def _listbox_filter_func(self, row):
         return True
 
     def _get_button_name_and_description(self, button):
+        # Translators: the {} will be replaced with the button index.
```
piotrdrag: What is a button index? Its number?

piotrdrag:
Please don’t make any changes to po/pl.po. I will take care of it.

Hjdskes:
> Please don’t make any changes to po/pl.po (other than re-generating it against a new .pot). I will take care of it.

Understood, I'll revert that part.

piper/buttondialog.py
Hjdskes: Yes; each button has an index, e.g. the left mouse button has index 0, the right mouse button has index 1, et cetera. If we map button 1 to button 0, the button with index 0 (left mouse button) will generate the click of button 1 (right mouse button).

piotrdrag: So how about:
`# Translators: the {} will be replaced with the button index, e.g. "Button 1 click".`

Hjdskes: Will do, thanks.

PR #158, opened 2017-08-24
ratbagd.py: add async signal for committed

### LINKED ISSUE #130: Make DBus methods async
See #119, depends on https://github.com/libratbag/libratbag/issues/269.

whot:
Fixes #130

Requires https://github.com/libratbag/libratbag/pull/296

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -186,6 +186,27 @@ class _RatbagdDBus(GObject.GObject):
         # update
         self._proxy.set_cached_property(property, val)
 
+    def _dbus_call_async(self, method, type, *value, callback):
+        # The callback handler to unwrap the data
+        def callback_handler(obj, result, user_callback):
+            res = obj.call_finish(result)
+            res = res.unpack()[0]  # Result is always a tuple
+            user_callback(res)
+
+        # Calls a method asynchronously on the bus, using the given method
+        # name, type signature and values. The callback is invoked when the
+        # method finishes
+        val = GLib.Variant("({})".format(type), value)
+        try:
+            self._proxy.call(method, val,
+                             Gio.DBusCallFlags.NO_AUTO_START,
+                             500, None, callback_handler, callback)
+        except GLib.Error as e:
+            # Unrecognized error code; print the message to stderr and raise
+            # the GLib.Error.
+            print(e.message, file=sys.stderr)
+            raise
+
     def _dbus_call(self, method, type, *value):
         # Calls a method synchronously on the bus, using the given method name,
         # type signature and values.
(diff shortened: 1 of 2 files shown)


whot:
Ok, re-pushed a new approach, this time in `ratbagd.py` only. I think there's some more exception handling needed but otherwise this works

piper/ratbagd.py
```diff
@@ -343,14 +364,16 @@ def get_svg(self, theme):
...
-    def commit(self):
-        """Commits all changes made to the device."""
-        ret = self._dbus_call("Commit", "")
+    def commit(self, callback):
+        """Commits all changes made to the device.
+
+        @param callback The function to call with the result of the commit
+        """
+        self._dbus_call_async("Commit", "", callback=callback)
         for profile in self._profiles:
```
Hjdskes: Shouldn't this also be done only if the commit succeeded? As it is right now, we might unflag a dirty profile while the commit fails in the background.

piper/ratbagd.py
```diff
@@ -186,6 +186,27 @@ def _set_dbus_property(self, property, type, value, readwrite=True):
...
+        val = GLib.Variant("({})".format(type), value)
+        try:
+            self._proxy.call(method, val,
+                             Gio.DBusCallFlags.NO_AUTO_START,
+                             500, None, callback_handler, callback)
+        except GLib.Error as e:
+            # Unrecognized error code; print the message to stderr and raise
+            # the GLib.Error.
+            print(e.message, file=sys.stderr)
+            raise
```
Hjdskes: Do we have an idea what errors can be raised here? It'd be better to try and catch them. The documentation lists the following but IIRC there are others too:

> If the `Gio.DBusConnection` associated with `self` is closed then the operation will fail with `Gio.IOErrorEnum.CLOSED`. If `cancellable` is canceled, the operation will fail with `Gio.IOErrorEnum.CANCELLED`. If parameters contains a value not compatible with the D-Bus protocol, the operation fails with `Gio.IOErrorEnum.INVALID_ARGUMENT`.

piper/ratbagd.py
whot: not with the plan of assuming commit always succeeds, see https://github.com/libratbag/libratbag/pull/275#discussion_r134724671

if commit fails, then we need to resync *everything* anyway and then we can also assume nothing is dirty.

Hjdskes: In that case, can we add a comment saying as such?

Also, I assume the resyncing will happen from within libratbag/ratbagd and that these changes will be signaled over the bus so that Piper picks that up automatically?

whot: > In that case, can we add a comment saying as such?

done

> Also, I assume the resyncing will happen from within libratbag/ratbagd and that these changes will be signaled over the bus so that Piper picks that up automatically?

yep, at least that's what ratbagd is supposed to do. if it doesn't, it's a bug there

whot:
@Hjdskes - ready for re-review

Hjdskes:
Pushed as a6c04d48026c7a3acfa8b460429b63dc9e1a29ac with some small changes.

PR #170, opened 2017-08-31
MouseMap: remove check for compliant SVGs

Hjdskes:
With the introduction of https://github.com/libratbag/libratbag/blob/master/data/gnome/check-svg.py, this check became obsolete.

### DIFF
--- a/piper/mousemap.py
+++ b/piper/mousemap.py
@@ -122,13 +122,6 @@ class MouseMap(Gtk.Container):
         self._handle = Rsvg.Handle.new_from_file(svg_path)
         self._svg_data = etree.parse(svg_path)
 
-        # TODO: remove this when we're out of the transition to toned down SVGs
-        device = self._handle.has_sub("#Device")
-        buttons = self._handle.has_sub("#Buttons")
-        leds = self._handle.has_sub("#LEDs")
-        if not device or not buttons or not leds:
-            print("Device SVG is incompatible", file=sys.stderr)
-
     def add(self, widget, svg_id):
         """Adds the given widget to the map, bound to the given SVG element
         identifier. If the element identifier or its leader is not found in the


whot:
hmm, not sure we should remove those. Because the SVGs are in two different repositories, we can't guarantee that the SVGs that piper sees are correct, so it should still do the checking. Otherwise you're just going to see exceptions etc. when you try to access the various bits and pieces later, right?

Might be worth moving the SVG checks into a separate function though and just check for everything we expect to be in the SVG.

Hjdskes:
> Might be worth moving the SVG checks into a separate function though and just check for everything we expect to be in the SVG.

Or we just run the `check_svg` script from Piper.. :sweat_smile: No but really, what are the quick run-time checks we can make from Piper? Checking for the three layers was a quick 'n dirty check while we were in the transition phase; if we're keeping this it should be something that can capture the essentials while not doing a full check like `check_svg` because it's run on every instantiation of a MouseMap widget.

bentiss:
I'd honestly go with the same script we are running in libratbag (or is it piper) that checks for the conformity of an SVG we ship. You never know if a user didn't put his own SVG in the data folder and then complain about a broken piper.

You should cache which SVG are opened globally, test those and remember it.

Hjdskes:
> You should cache which SVG are opened globally, test those and remember it.

Interesting approach, but where would I store this? GSettings comes to mind because it's persistent across executions of Piper and I'd prefer that over writing to a random file in a user's `$HOME`.

bentiss:
> Interesting approach, but where would I store this? GSettings comes to mind because it's persistent across executions of Piper and I'd prefer that over writing to a random file in a user's $HOME.

I don't think you should cache this across executions of piper. If a user has a faulty SVG, it's likely that he will rework on the design and then re-run piper.
You should IMO cache it in a global dict (or in the Ratbagd class, or in any class you think would fit). 

Actually, you should probably simply check for the validity of the SVG the first time you open it. This way you can have:
- user puts a first design in data folder
- user run piper, which complains
- user fix the design
- user unplug/re-plug the mouse
- piper is now happy without even having to restart it

Hjdskes:
> I don't think you should cache this across executions of piper. If a user has a faulty SVG, it's likely that he will rework on the design and then re-run piper.

I'd only cache known working SVGs in this case, but this doesn't catch the instances where a user edits an SVG that was previously known to be good so indeed, let's not go across executions.

> You should IMO cache it in a global dict (or in the Ratbagd class, or in any class you think would fit).

This could fit in the `Application` class, in which case we can cache it even over different windows of Piper. I'll play around for a bit to see what I can come up with.

whot:
the vast majority of our users will have one, maybe two devices and the only use-case for broken SVGs is when we add new ones or when something really breaks (user changing things or piper updating in incompatible ways). I'm not sure caching an SVG is really worth it here.

How about changing the check_svg.py into something that's usable from code instead of just being a standalone script and running that when loading the svg?

Issue #177, opened 2017-09-15
Crash on applying any settings


joonatoona:
```
Ubuntu 17.04
Piper from git at commit 4f560b0d1e13383a075dfe94de106942a2028860
libratbag from git at commit 52e10d029f7cc7573f55b63dc435e706c021d72d
Logitech G502 Proteus Spectrum
```

```
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/piper/ratbagd.py", line 192, in callback_handler
    res = obj.call_finish(result)
GLib.Error: g-io-error-quark: Timeout was reached (24)
```

Hjdskes:
The very least we should do here is catch those exceptions so they don't bring Piper down, but ideally we need to figure out why the timeout is occurring in the first place.

I can't find what a timeout means in this asynchronous context. Do you know @whot?
We can pass `GLib.G_MAXINT` to disable it if it doesn't make sense, otherwise we'll have to look into enlarging it as well. For synchronous calls the timeout is 2 seconds, for asynchronous it currently is half a second.

muzena:
@joonatoona 
Can you start piper in Ubuntu 17.04?

joonatoona:
What do you mean? I am in Ubuntu 17.04?

muzena:
@joonatoona 
I have this problem in Ubuntu 17.10:
https://github.com/libratbag/piper/issues/66

joonatoona:
@muzena 

```env PYTHONPATH=$PYTHONPATH:/usr/lib/python3.5/site-packages/ piper```

seems to work just fine

muzena:
Ok, I will try

muzena:
@joonatoona 
I'm trying to compile on clean install Ubuntu 17.04?
I compile and install libratbag
then I try compile ratbag, and tell me that missing libratbag dependencies,
then I install manually libratbag deb packages from   Ubuntu 17.10

compilation gave me this error:
../src/ratbagd-device.c: In function ‘ratbagd_device_get_capabilities’:
../src/ratbagd-device.c:283:3: error: ‘RATBAG_DEVICE_CAP_RESOLUTION’ undeclared (first use in this function)
   RATBAG_DEVICE_CAP_RESOLUTION,
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/ratbagd-device.c:283:3: note: each undeclared identifier is reported only once for each function it appears in
[...]

PR #178, opened 2017-09-16
ratbagd: move catching exception to the right place

Hjdskes:
See #177; [`Gio.DBusProxy::call`](https://lazka.github.io/pgi-docs/Gio-2.0/classes/DBusProxy.html#Gio.DBusProxy.call) does not raise exceptions, but [`Gio.DBusProxy::call_finish`](https://lazka.github.io/pgi-docs/Gio-2.0/classes/DBusProxy.html#Gio.DBusProxy.call_finish) does. For starters, we should thus move our catching there.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -189,23 +189,23 @@ class _RatbagdDBus(GObject.GObject):
     def _dbus_call_async(self, method, type, *value, callback):
         # The callback handler to unwrap the data
         def callback_handler(obj, result, user_callback):
-            res = obj.call_finish(result)
-            res = res.unpack()[0]  # Result is always a tuple
-            user_callback(res)
+            try:
+                res = obj.call_finish(result)
+                res = res.unpack()[0]  # Result is always a tuple
+                user_callback(res)
+            except GLib.Error as e:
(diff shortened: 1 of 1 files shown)


Hjdskes:
I would like to rework this exception handling in ratbagd; since https://github.com/libratbag/libratbag/pull/336 we don't return ratbag error values over DBus anymore so trying to detect these in `_dbus_call` is moot. This then brings us back to where the only exceptions that occur over the bus are DBus errors (think timeouts for one), so that makes things a little easier for us there. The question then is: what do we do once we catch such an exception? Up until now we've always just re-raised the exception, but I'd like to get this fixed.

One thing that might work is to emit a signal that Piper (or any client) can connect to, which indicates something went wrong. Piper can then inform the user of this using its in-app notification. What do you think @whot?

whot:
tbh, I think what this does is sufficient for now and could be merged. The signal approach sounds sensible too, the only other way I can come up with is to have commit take an (optional) error_handler callback that's called with the exception. But we can do that when we have real need for it.

Hjdskes:
I'm merging this because I don't have the time to implement the signal at the moment, but that shouldn't hold back this fix.

PR #183, opened 2017-09-29
ratbagd: sync with libratbag

Hjdskes:


whot:
sorry, somehow I never saw this one but I think this is fixed as part of #186

PR #186, opened 2017-10-15
Update to new libratbag API

Hjdskes:


phomes:
Thanks. I tested and Piper works for me again. It does not restrict the slider to the new DPI list, but only uses the min/max values. Better than crashing though :)

Hjdskes:
I don't have that much time atm so preventing the crash is all I can do for now. In three or four weeks I should be able to invest more time in Piper again!

Meanwhile, restricting values shouldn't be that hard... ;)

phomes:
> Meanwhile, restricting values shouldn't be that hard... ;)

So you will do it? Or was that a hint for me to do so?

whot:
thx!

PR #187, opened 2017-10-16
ResolutionsRow: snap to nearest allowed resolution value

whot:
libratbag now exports a list of permitted resolutions rather than a min/max
value (and a guesswork step increment on piper's side). Use that and snap the
resolution slider to the nearest value in that list.

See https://github.com/libratbag/libratbag/pull/393

Related to #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,11 @@ README](https://github.com/libratbag/libratbag/blob/master/README.md#running-rat
 Screenshots
 ===========
 
-![resolution configuration screenshot](https://github.com/whot/piper/blob/wiki/screenshots/piper-resolutionpage.png)
+![resolution configuration screenshot](https://github.com/libratbag/piper/blob/wiki/screenshots/piper-resolutionpage.png)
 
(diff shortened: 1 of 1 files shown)


phomes:
Both patches worked well for me

Hjdskes:
I'll review this ASAP, but it might be a few days.

phomes:
the only commit that was merged here was for the screenshots in the README. Was that intentional?

whot:
@phomes: yeah, that one got included here by mistake. The other two are still waiting for review, though I think I'll merge it in a few days if @Hjdskes doesn't get to it. We can fix it later if need be.

phomes:
@whot yes, but now this PR is closed (merged) and the only commit shown here is one for the README. So maybe a new PR for the other two commits is needed?

whot:
oops, I didn't notice that. I can't remember what the other two commits were, but I pushed #194 now

Hjdskes:
Thanks, sorry for the delay.

PR #188, opened 2017-10-16
Led: add an image for the led off page

### LINKED ISSUE #45: The Off page in the LED dialog needs an image
It needs an image according to the [empty states](https://developer.gnome.org/hig/stable/empty-placeholder.html.en) guidelines, see also https://wiki.gnome.org/Design/OS/EmptyStates

whot:
Ah, what a beautiful night sky...

The CSS background taken from the control-center scroll test page. Apparently
we can't use the SVG background and have to do this manually.

Fixes #45

### DIFF
--- a/piper/leddialog.py
+++ b/piper/leddialog.py
@@ -34,6 +34,7 @@ class LedDialog(Gtk.Dialog):
     colorbutton = GtkTemplate.Child()
     adjustment_brightness = GtkTemplate.Child()
     adjustment_effect_rate = GtkTemplate.Child()
+    led_off_image = GtkTemplate.Child()
 
     def __init__(self, ratbagd_led, *args, **kwargs):
         """Instantiates a new LedDialog.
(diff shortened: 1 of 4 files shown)


piper/leddialog.py
```diff
@@ -60,6 +61,11 @@ def __init__(self, ratbagd_led, *args, **kwargs):
         self.adjustment_brightness.set_value(self._led.brightness)
         self.adjustment_effect_rate.set_value(self._led.effect_rate)
 
+        sp = Gtk.CssProvider()
+        sp.load_from_data("* { background: #565854}".encode())
```
Hjdskes: In general I don't like hardcoding colors because this may break completely with different themes. Isn't there a color we can retrieve from the theme (through https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-color) such that it will work for all themes? This is the approach used in the mousemap as well.

whot: looks like that function is deprecated, so we shouldn't use it. it directs me to [gtk_render_background](https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-render-background) instead, but that suggests its based on CSS data anyway? That magic value is taken from the gnome-control-center, so that too would break with nondefault themes.

Hjdskes: > looks like that function is deprecated

Where do you see that?

> it directs me to `gtk_render_background` instead, but that suggests its based on CSS data anyway?

CSS data taken from whatever theme the user has applied, yes. Not a magic value taken from gnome-control-center that only works with Adwaita. At least, if I understand it correctly.

Hjdskes: >> looks like that function is deprecated
>Where do you see that?

Oh, oops. I just see I've been looking at the foreground color version and not the background color version, which is indeed deprecated. `gtk_render_background` seems to be the wrong function as I think it has to be used in the `draw` function only.

In that case, let's go with this and revisit if people start complaining about custom themes breaking.

PR #192, opened 2017-10-19
Do not generate icon-cache when DESTDIR is set

ikeydoherty:
This ensures that the icon cache will be generated when installed locally,
but will also make sure that we don't generate one when DESTDIR is set.
This helps with packaging, where packages should never provide the duplicate
"icon-theme.cache" file inside the package.

This is based around the meson build system implemented by @TingPing in
the gnome-twitch project.

Signed-off-by: Ikey Doherty <ikey@solus-project.com>

### DIFF
--- a/meson_install.sh
+++ b/meson_install.sh
@@ -1,7 +1,11 @@
 #!/usr/bin/env sh
 
-# Update icon cache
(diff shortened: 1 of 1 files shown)


Hjdskes:
Learned something new, thanks! :)

ikeydoherty:
My pleasure - we're really loving the libratbag projects in Solus, happy to help :)

PR #194, opened 2017-10-25
ResolutionsRow: snap to nearest allowed resolution value

whot:
libratbag now exports a list of permitted resolutions rather than a min/max
value (and a guesswork step increment on piper's side). Use that and snap the
resolution slider to the nearest value in that list.

See https://github.com/libratbag/libratbag/pull/393

Related to #186


whot:
ok, this is #187 force-pushed again. github doesn't like changing commits...

PR #195, opened 2017-10-29
MouseMap: if the old highlight is None, don't try to redraw

### LINKED ISSUE #185: Crash when dragging a button
I accidentally clicked and dragged one of the buttons to configure a LED. Piper crashes with the following:

```
  File "/home/phomes/piper/piper/mousemap.py", line 318, in _on_leave
    self._redraw_svg_element(old_highlight)
  File "/home/phomes/piper/piper/mousemap.py", line 362, in _redraw_svg_element
    ok, svg_geom = self._get_svg_sub_geometry(svg_id)
  File "/home/phomes/piper/piper/mousemap.py", line 340, in _get_svg_sub_geometry
    ok, svg_pos = self._handle.get_position_sub(svg_id)
TypeError: Argument 2 does not allow None as a value

```

whot:
When pressing a button and moving out of the button area with the button held
down, we get two leave events - once when actually leaving the button and once
more when the button press is released later. On the second call, the old
highlight is already None, causing a crash when we're trying to get SVG
information about it.

Fixes #185

### DIFF
--- a/piper/mousemap.py
+++ b/piper/mousemap.py
@@ -314,6 +314,10 @@ class MouseMap(Gtk.Container):
     def _on_leave(self, widget, event):
         # Restores the device SVG to its original state.
         old_highlight = self._highlight_element
+
+        if old_highlight is None:
+            return
+
         self._highlight_element = None
         self._redraw_svg_element(old_highlight)
 


phomes:
CI failure seems unrelated

PR #196, opened 2017-11-11
LedDialog: use new effect duration in ms

### LINKED ISSUE #175: 100Hz LED as minimum is much to fast
Setting the minimum frequency of a 100Hz for 'Breathing' or 'Cycling' is somewhat excessive as it results in a strobe effect.

Also when I apply changes to the LEDs the program crashes with:
File "/usr/lib/python3.6/site-packages/piper/ratbagd.py", line 192, in callback_handler
res = obj.call_finish(result)
GLib.Error: g-io-error-quark: Timeout was reached (24)

After the crash and restarting piper the sensor refresh rate is set to 500Hz while it was set to a 1000Hz before. After each crash it defaults back to 500Hz.

Hjdskes:
libratbag/libratbag#422 introduced effect duration in ms for LEDs as opposed to effect rate in Hz, as was discussed in #175. This commit updates Piper's UI to reflect these changes.

Fixes #175.

NOTE: this depends on libratbag/libratbag#422 and should not be changed before 1) that PR is merged and 2) its changes to ratbagd.py have been synced back into Piper.

### DIFF
--- a/data/ui/LedDialog.ui
+++ b/data/ui/LedDialog.ui
@@ -7,11 +7,11 @@
     <property name="step_increment">1</property>
     <property name="page_increment">10</property>
   </object>
-  <object class="GtkAdjustment" id="adjustment_effect_rate">
-    <property name="lower">100</property>
-    <property name="upper">20000</property>
-    <property name="step_increment">50</property>
-    <property name="page_increment">50</property>
+  <object class="GtkAdjustment" id="adjustment_effect_duration">
+    <property name="lower">0</property>
+    <property name="upper">10000</property>
+    <property name="step_increment">100</property>
+    <property name="page_increment">500</property>
   </object>
   <template class="LedDialog" parent="GtkDialog">
     <property name="width_request">400</property>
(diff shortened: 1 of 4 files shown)


data/ui/LedDialog.ui
```diff
@@ -7,11 +7,11 @@
...
-  <object class="GtkAdjustment" id="adjustment_effect_rate">
-    <property name="lower">100</property>
-    <property name="upper">20000</property>
-    <property name="step_increment">50</property>
-    <property name="page_increment">50</property>
+  <object class="GtkAdjustment" id="adjustment_effect_duration">
+    <property name="lower">0</property>
+    <property name="upper">10000</property>
+    <property name="step_increment">500</property>
+    <property name="page_increment">500</property>
```
whot: half-second increments seem a bit high. Maybe make it at 100 for step and 500 for page?

Hjdskes:
Alright, everything is ready to go here. @whot, can you take a last look and merge if everything is alright?

PR #213, opened 2018-07-03
meson: check if libratbag is installed (via ratbagd)

FFY00:


Hjdskes:
Well, it's certainly working! :smile: 

Can you add a separate commit that installs ratbagd in the CI build, to see if it then passes?

meson.build
```diff
@@ -6,6 +6,11 @@ project('piper',
 # Dependencies
 dependency('python3', required: true)
 dependency('pygobject-3.0', required: true)
+ratbagd = find_program('ratbagd', required: false)
+
+if not ratbagd.found()
```
Hjdskes: Doesn't this do exactly the same as setting `required: true` in the `find_program` directive?

FFY00:
The message shown isn't clear to me. It says the program `ratbagd` wasn't found. Anyway, it's your call.

FFY00:
The CI need to be fixed. I don't use fedora so I don't know how to fix it. I tried to add bentiss' copr repository and install it but that didn't work.

whot:
Urgh, turns out I haven't updated Fedora 28 with the current ratbagd, so we'll have to fix circleci first here, sorry.

whot:
Just FYI, currently waiting on libratbag to end up in F28 stable so we can rely on it in circleci

FFY00:
Why don't we use Arch Linux instead of Fedora. The package is almost always up to date and if it isn't I can do something about it. We could even make our own repository if needed.

You can check an example CircleCI file here: https://gergely.imreh.net/blog/2018/04/circleci-aur

whot:
> Why don't we use Arch Linux instead of Fedora?

No specific reason other than that @bentiss and I are the Fedora maintainers, so we can always change it where needed. So same reasoning you have :) Wasn't a problem so far, but with the bump to merge ratbagd and all the dbus changes I first avoided pushing the breakage to fedora, then I forgot about it. Switching to Arch works for me though.

whot:
Can you rebase and force-push this branch please? ratbagd is now in f28 stable so this time the checks should succeed, I can then merge the branch. Thanks

whot:
Nevermind, I realised I can just take the commit and push it out myself ;) Coffee is an amazing thing...

Pushed out as part of https://github.com/libratbag/piper/pull/218, commit 132da58911449770ce66f494b696ba311e269c68

PR #229, opened 2018-08-07
ratbagd: enums are sent as signed int

phomes:
ratbagd was updated to send enums as signed int values in
libratbag/libratbag/pull/504
This updates Piper to do the same


Hjdskes:
Just making sure: these changes should also be synced to libratbag again.

phomes:
> Just making sure: these changes should also be synced to libratbag again.

I updated the patch for libratbag to also have these updates for ratbagctl. Thanks

whot:
I think with libratbag commit  447931d7791abdfd and 447931d7791abdfd we can now close this.

PR #230, opened 2018-08-07
Use profile name from libratbag

phomes:
libratbag stores a name for each profile since
libratbag/libratbag/pull/275

Let's use the supplied name instead of using index numbers


whot:
test case failure is a flake8 check

```./piper/profilerow.py:17:1: F401 'gettext.gettext as _' imported but unused```

Can you update your branch to make that go away and merge it? Thanks

Issue #5, opened 2017-05-15
Create a Flatpak including libratbag and ratbagd


Hjdskes:
Hello,

@whot asked me to look into creating a Flatpak of Piper which includes libratbag and ratbagd. This issue is meant to track required items. I will edit this post to add items as I go; other findings will be added as comments.

RUNTIME: we will require GNOME's runtime as we require GLib and GTK+ for Piper. Versions generally don't matter, but we need to see what version of GTK+ we want to use. Newer versions have fancier widgets, but this might limit availability in distributions like Debian when Piper is installed without Flatpak.

TODO:
- [ ] Make ~~libratbag, ratbagd and~~ Piper build systems compliant with the [Build API](https://github.com/cgwalters/build-api/) as required for `flatpak-builder`
  - [x] libratbag
  - [x] ratbagd
  - [ ] Piper
- [ ] AppData file for Piper (optional)
- [ ] All files exported from the Flatpak need to have the application ID as their prefix. We can either rename the relevant source files or use `flatpak-builder` to rename the files at build time. My preference would be the former to keep it simple.
  - [ ] Desktop file
  - [ ] Application icon
  - [ ] `ratbagd.service.in` (not sure how this works with an `org.freedesktop.ratbagd` ID in the Flatpak for Piper with `org.freedesktop.Piper` yet)

Hjdskes:
Just found out that [Meson is supported natively by `flatpak-builder`](https://github.com/flatpak/flatpak/pull/420) so we don't need to touch libratbag and ratbagd's build systems.

whot:
but we don't need to *build* anything for a pure python application, don't we? We just need to *run* it, and that should be possible without any extra work, it can be treated like the hello world example application: http://flatpak.org/hello-world.html

I think so, anyway :)

Hjdskes:
Right, I thought it'd be best to stick to executing `setup.py` in the Flatpak but simply copying files over should work too. I've begun building with the former, though, as it seems this can then be automated with `flatpak-builder` whereas the hello-world example seems to imply it's manual work.

I've run into another, nastier issue, though:
```
Meson encountered an error in file meson.build, line 44, column 0:
Native dependency 'libudev' not found
```

[...]

PR #235, opened 2018-08-22
Flatpak support

### LINKED ISSUE #202: Reinvestigate a Piper Flatpak
As mentioned in #199, it may now be possible (although perhaps too early still) to Flatpak Piper. See #5 for the previous discussion.

whot:
Fixes #202, #236 

Edit: still missing
- [x] desktop icon
- [x] appdata
- [x] `.desktop` file

### DIFF
--- a/org.freedesktop.Piper.json
+++ b/org.freedesktop.Piper.json
@@ -1,0 +1,63 @@
+{
+    "app-id": "org.freedesktop.Piper",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.28",
+    "sdk": "org.gnome.Sdk",
+    "command": "piper",
+      "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
+        "--filesystem=xdg-run/dconf",
+        "--filesystem=~/.config/dconf:ro",
+        "--talk-name=ca.desrt.dconf",
+        "--env=DCONF_USER_CONFIG_DIR=.config/dconf",
+        "--system-talk-name=org.freedesktop.ratbag1"
+      ],
+    "modules": [
+        {
+            "name": "python-evdev",
+            "buildsystem": "simple",
+            "build-commands": [
+                "pip3 install --prefix=/app --no-deps ."
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://files.pythonhosted.org/packages/da/c9/8f88a77b6f934d60e3ff0e210ed7a15f47a2e9c05e4c8fea67d99d670ab1/evdev-1.0.0.tar.gz",
+                    "sha256": "be0288ef1125bf1c539bb8f3079ef4aa5fb813af28f0c5294a4e744ee554398a"
+                }
+            ]
+        },
+        {
+            "name": "python-lxml",
+            "buildsystem": "simple",
+            "build-commands": [
+                "pip3 install --prefix=/app --no-deps ."
+            ],
+            "build-options": {
+                "env": {
+                    "XSLT_CONFIG": "pkg-config libxslt"
+                }
+            },
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://files.pythonhosted.org/packages/ca/63/139b710671c1655aed3b20c1e6776118c62e9f9311152f4c6031e12a0554/lxml-4.2.4.tar.gz",
+                    "sha256": "e2afbe403090f5893e254958d02875e0732975e73c4c0cdd33c1f009a61963ca"
+                }
+            ]
+        },
+        {
+            "name": "Piper",
+            "buildsystem": "meson",
+            "config-opts": [ "-Dignore-ratbagd-dependency=true" ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/libratbag/piper"
+                }
+            ]
+        }
+    ]
+}
--- a/data/meson.build
+++ b/data/meson.build
@@ -2,10 +2,12 @@ gnome = import('gnome')
 
 desktopdir = join_paths(datadir, 'applications')
 icondir = join_paths(datadir, 'icons', 'hicolor', 'scalable', 'apps')
+metainfodir = join_paths(datadir, 'metainfo')
 
 conf = configuration_data()
 conf.set('version', meson.project_version())
 conf.set('url', 'https://github.com/libratbag/piper')
+conf.set('version_date', version_date)
 
 about_dialog = configure_file(input: 'ui/AboutDialog.ui.in',
 			      output: 'AboutDialog.ui',
@@ -20,3 +22,8 @@ gnome.compile_resources('piper', 'piper.gresource.xml',
 
 install_data('org.freedesktop.Piper.desktop', install_dir: desktopdir)
 install_data('org.freedesktop.Piper.svg', install_dir: icondir)
+
+appdata = configure_file(input: 'org.freedesktop.Piper.appdata.xml.in',
+                         output: 'org.freedesktop.Piper.appdata.xml',
+                         configuration: conf,
+                         install_dir: metainfodir)
--- a/data/org.freedesktop.Piper.appdata.xml.in
+++ b/data/org.freedesktop.Piper.appdata.xml.in
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.freedesktop.Piper</id>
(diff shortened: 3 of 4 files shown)


mati865:
It doesn't ship `ratbagd`, is there runtime notification if it has to be installed?

org.freedesktop.Piper.json
```diff
@@ -0,0 +1,60 @@
...
+    "app-id": "org.freedesktop.Piper",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.28",
+    "sdk": "org.gnome.Sdk",
+    "command": "piper",
+      "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
+        "--system-talk-name=org.freedesktop.ratbag1"
```
Hjdskes: Is the idea now to have ratbagd running in the system and allow the sandboxed Piper to communicate with it over DBus through `"--system-talk-name=org.freedesktop.ratbag1"`? This doesn't seem to be working for me at the moment.

Otherwise, with libratbag still requiring udev, aren't we in the same boat we were in in https://github.com/libratbag/piper/issues/5?

Hjdskes: Is the idea now to have ratbagd running in the system and allow the sandboxed Piper to communicate with it over DBus through `"--system-talk-name=org.freedesktop.ratbag1"`? This doesn't seem to be working for me at the moment.

Otherwise, with libratbag still requiring udev, aren't we in the same boat we were in in https://github.com/libratbag/piper/issues/5?

Hjdskes:
For other testers, change the following to build the flatpak:

```diff
-                    "url": "https://github.com/libratbag/piper"
+                    "url": "https://github.com/whot/piper",
+                    "branch": "wip/flatpak"
```

whot:
> It doesn't ship ratbagd, is there runtime notification if it has to be installed?

piper has always displayed a "could not connect to ratbagd" message anyway, the meson.build addition was to avoid this happening to almost everyone who builds piper. This is a bit different with flatpak - we'd rely on the system daemon to be present but at least we can provide an up-to-date piper without relying on ppl to build piper from scratch.

> I only get an accessibility warning:

That one and I get ``Gtk-Message: 08:08:54.393: Failed to load module "pk-gtk-module"`` too. I'll look into both, but wanted to get the initial support out yesterday before going to bed.

org.freedesktop.Piper.json
Hjdskes: I think D-Feet showed the regular dbus name, but I'll look into this tomorrow.

whot:
So, turns out that the icon and the `.desktop` file seems to work just fine once the flatpak is installed and run from the installed version. Only when run with `flatpak-builder` does it get an empty icon.

whot:
Moved the appstream data requirement to #236 since it's not really part of the flatpak support itself, we can do this independently.

julianrichen:
I don't know the technical layout of libratbag but I wanted to offer a suggestion for the Piper flatpak. I was working on a Piper flatpak myself but, as mentioned above one of the limitation was packaging ratbagd in the flatpak. Flatpak does allow exposing `/dev`, but flatpak doesn't currently offer a solution for running a daemon & application at the same time. Piper would need to launch the daemon (I believe `org.gnome.FeedReader` does this). Plus meeting the systemd requirement for udev was a pain.

One of the reasons for packaging libratbag was the updated support for new devices & drivers. Instead of packaging libratbag completely could the Piper flatpak have extensions point for a development/updated drivers and devices? Either a generic `org.freedesktop.Piper.Drivers` or per driver extension `org.freedesktop.Piper.Drivers.hidpp20`. Same for `org.freedesktop.Piper.Devices` or `org.freedesktop.Piper.Devices.g203`.

Looking at the libratbag code it seems drivers are currently compiled in ratbagd. Could they be split and picked-up from a default folder (`/lib64/ratbag/drivers`) and a specified directory (eg. `/app/lib/ratbag/drivers/`). The flatpak package wouldn't required the `*.Drivers` extension and would use the default drivers, however, if the extension is found they offer the latest drivers. ratbag would also need a way to pick-up drivers automatically when they become available (ie. when the flatpak is launched).

I imagine this is a fair amount of work to-do in libratbag, so might be something worth looking into later.

whot:
this is a bit like putting the card before the horse. The drivers are an integral part of libratbag with a non-stable API between libratbag and the drivers. Splitting them out would be yet another API that we cannot easily change on-demand. Piper itself is driver-agnostic and only requires dbus API knowledge so shipping the drivers with piper doesn't make much sense. FTR, piper is only one possible GUI to libratbag, it'd be trivial to write another one for e.g. KDE. They are all driver agnostic then.

There's also the problem that to expose write access to /dev, the flatpak must run as root (afaict) and that's no good for piper. Shipping libratbag as flatpak is theoretically doable but requires some acrobatics, see #5.

So yeah, right now piper and libratbag will have to stay separate.

whot:
Force pushed after a rebase and we now have appdata too (see #236). This should be ready to merge now.

Hjdskes:
I'll try to review this today, sorry for the delay.

Hjdskes:
Please cherry-pick https://github.com/Hjdskes/piper/commit/211bba5996e4726d71e1ee83a4f6289c0908eb11, which adds some missing bits to the AppData file:

* a content rating, needed if we want to publish Piper on Flathub;
* some kudos, which will improve the rating of Piper in GNOME Software;
* a help URL and a proper dated release version

Without at least a dated release, `flatpak-builder` does not want to build the Flatpak because the AppData file is invalid.

Building and running the Piper Flatpak with the following instructions makes it work properly. I have the proper icon and the dbus warning is gone. Ratbagd cannot find my device at the moment but that is unrelated to the Flatpak, I suppose. I do still get a warning that the locale is not supported by the C library, but I have a mixed en_US and nl_NL locale which might be the reason.

```
flatpak-builder --force-clean --repo=repo piper org.freedesktop.Piper.json
flatpak --user remote-add --no-gpg-verify --if-not-exists piper repo
flatpak --user install piper org.freedesktop.Piper

flatpak run org.freedesktop.Piper
```

Will we publish Piper on Flathub? It's straightforward to do; I can do it this weekend if we want it.

Hjdskes:
According to [the documentation](http://docs.flatpak.org/en/latest/sandbox-permissions.html), adding `--filesystem="/usr/share/libratbag/gnome:ro"` should provide read only access to this directory, but it doesn't seem to be working in my tests.

whot:
> a help URL and a proper dated release version

fwiw, I changed things a bit in the 8db55d2 commit, there was a copy/paste in your commit and it now points at the piper wiki. Also shuffled the date around so it's set in meson.build, easier to (not forget to) update this way.

Hjdskes:
I meant the help url to point to the issues page, as the wiki is quite empty at the moment. But this works too.

Regarding the version, I think the point is to also keep listing older versions. Using meson to fill this in means we only list te current release.

Hjdskes:
> right, but your commit was pointing to some completely different repo :)

Oh, whoops :)

org.freedesktop.Piper.json
```diff
@@ -0,0 +1,60 @@
+{
+    "app-id": "org.freedesktop.Piper",
+    "runtime": "org.gnome.Platform",
+    "runtime-version": "3.28",
+    "sdk": "org.gnome.Sdk",
+    "command": "piper",
+      "finish-args": [
+        "--share=ipc",
+        "--socket=x11",
+        "--socket=wayland",
```
Hjdskes: This requires access to DConf, see https://github.com/flatpak/flatpak-docs/blob/master/docs/sandbox-permissions.rst#dconf-access. I learned this when submitting my own application to Flathub: https://github.com/flathub/flathub/pull/601#discussion_r214522390

org.freedesktop.Piper.json
```diff
@@ -0,0 +1,60 @@
...
+                    "type": "archive",
+                    "url": "https://files.pythonhosted.org/packages/ca/63/139b710671c1655aed3b20c1e6776118c62e9f9311152f4c6031e12a0554/lxml-4.2.4.tar.gz",
+                    "sha256": "e2afbe403090f5893e254958d02875e0732975e73c4c0cdd33c1f009a61963ca"
+                }
+            ]
+        },
+        {
+            "name": "Piper",
+            "buildsystem": "meson",
+            "builddir" : true,
```
Hjdskes: Can remove this, it is forced for Meson: https://github.com/flathub/flathub/pull/601#discussion_r214521469

whot:
dconf added, builddir line removed as requested.

PR #246, opened 2018-08-30
Fetch SVGs as fds, not as paths.

whot:
Required for #235, otherwise we can't open the files. Requires https://github.com/libratbag/libratbag/pull/523 to be pulled in first.

The flatpak branch in #235 needs to be rebased on top of this one once merged, then it should all just magically work.

### DIFF
--- a/piper/mousemap.py
+++ b/piper/mousemap.py
@@ -16,14 +16,13 @@
 
 import cairo
 import gi
-import os
 import sys
 from lxml import etree
 
 gi.require_version("Gdk", "3.0")
 gi.require_version("Gtk", "3.0")
 gi.require_version("Rsvg", "2.0")
-from gi.repository import Gdk, GLib, Gtk, GObject, Rsvg
+from gi.repository import Gdk, GLib, Gtk, GObject, Rsvg, Gio
 
 """This module contains the MouseMap widget (and its helper class
 _MouseMapChild), which is central to the button and LED configuration stack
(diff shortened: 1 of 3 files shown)


piper/mousemap.py
```diff
@@ -119,8 +119,13 @@ def __init__(self, layer, ratbagd_device, spacing=10, *args, **kwargs):
...
         self._highlight_element = None
 
-        self._handle = Rsvg.Handle.new_from_file(svg_path)
-        self._svg_data = etree.parse(svg_path)
+        uis = Gio.UnixInputStream.new(fd.fileno(), False)
+        self._handle = Rsvg.Handle.new_from_stream_sync(uis, None,
+                                                        Rsvg.HandleFlags.FLAGS_NONE,
+                                                        None)
+        # Same fd, so let's reset it before we parse it
+        fd.seek(0)
```
Hjdskes: Should we perhaps reset it after reading it in the `DeviceRow`, and again here? I don't know how costly this is, but it is more "future proof" to just reset it every time we read it.

piper/ratbagd.py
```diff
@@ -376,7 +405,8 @@ def get_svg(self, theme):
 
         @param theme The theme from which to retrieve the SVG, as str
         """
-        return self._dbus_call("GetSvg", "s", theme)
+        fd = self._dbus_call_return_one_fd("GetSvg", "s", theme)
```
Hjdskes: For Piper at least, we could also do all the reading here and return a RSVG handle here. That would move all the reading to a single place, where we can reset the fd and remove the duplication now present in the device row and the mouse map. This is my preferred solution.

How would this work for the bindings for ratbagctl?

whot: well, we don't know in the bindings what the SVG is being used for. Even in piper we use it as both SVG and XML file (hence the need to `seek(0)`).  I have this fixed as simply calling the DBus API twice now, this way it resembles the original code (where we opened the file twice too)

whot: Oh, and: to avoid the duplication we'd need some util function in some module, etc. All ends up being more complicated than two lines of duplicated code, so I don't think it's worth it

Hjdskes: Right, I forgot about the XML use case.

whot:
urgh, copy/paste error, forgot the `fileno()` in the devicerow. Fixed now

whot:
557a1a8 is just a rebase on top of current piper master

PR #248, opened 2018-09-04
meson.build: add hooks to check for a minimum ratbagd version

whot:
Checks the ratbagd version at meson time and expresses a warning when it's not recent enough.

Up until the currently requested minimum version, ratbagd didn't have a
--version command so an error is equivalent to 'too old'.


whot:
libratbag 0.9.903 is out now, so we can merge this

PR #251, opened 2018-09-10
data: link to flathub-compatible 16:9 screenshots in the appdata

whot:
Fixes: https://github.com/flathub/org.freedesktop.Piper/issues/1

Pretty sure that's the reason we don't show up in flathub.


Hjdskes:
I do think the screenshots can look nicer by centering the mousemap vertically (it's all about advertisement), but that is of secondary importance at the moment.

PR #259, opened 2018-09-12
Use Unicode typography in user-visible strings

piotrdrag:
See https://developer.gnome.org/hig/stable/typography.html


Hjdskes:
Thanks!

PR #274, opened 2018-11-27
Handle daemon crashes/disappearance

whot:
Let's not crash when the daemon goes away

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -278,12 +278,18 @@ class Ratbagd(_RatbagdDBus):
             (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
         "device-removed":
             (GObject.SignalFlags.RUN_FIRST, None, (GObject.TYPE_PYOBJECT,)),
+        "daemon-disappeared":
+            (GObject.SignalFlags.RUN_FIRST, None, ()),
     }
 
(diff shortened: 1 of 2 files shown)


whot:
CI check failure is due to the now-fixed #276, please ignore.

Hjdskes:
Should there also be some logic in case ratbagd is restarted by the user while Piper is running? In any case, I'm merging this for now.

whot:
There should be, but whether it's worth it is another question. In a good setup, ratbagd will start once piper tries to access it over dbus. If it crashes, we don't need piper to wait around.

And with the timeout to shut down ratbagd after 20 min - well, you can restart piper if you haven't done anything for 20 min :)

Hjdskes:
I just looked into this a bit more. For reference, my point was that if the user is using Piper and ratbagd crashes, they might open a terminal and manually restart ratbagd while Piper remains open in the background. In such cases, it would have been nice for Piper to pick up on this and switch back to the mouse- or welcome perspective.

However, now that I'm looking into it, that's a lot of work for what is likely an edge case (and even if it occurs, I think most users are more likely to simply restart Piper). The problem namely is, that when the name is acquired again on the bus, we'd need to update all `ratbag` references throughout all Piper objects. That is simply not worth it.

whot:
yeah, I agree. thanks for looking into it though.

PR #275, opened 2018-11-27
mouseperspective: don't show the profile button if we only have one profile

### LINKED ISSUE #270: Don't show the profile button if we don't have a profile
On the G500s the button is still visible, without content so it's just a square block. Clicking on it shows an empty profle and the Add Profile, but greyed out.

If we don't have switchable profiles anyway, we shouldn't show that button to begin with.

whot:
Fixes #270


whot:
CI check failure is due to the now-fixed #276, please ignore.

whot:
Thanks!

PR #279, opened 2018-12-11
Remove async

Hjdskes:
Since libratbag/libratbag#563, the implementation in ratbagd is made async. Hence, we do not need the
asynchronisity provided by DBus anymore.

### DIFF
--- a/piper/mouseperspective.py
+++ b/piper/mouseperspective.py
@@ -19,7 +19,7 @@ from gettext import gettext as _
 from .buttonspage import ButtonsPage
 from .gi_composites import GtkTemplate
 from .profilerow import ProfileRow
-from .ratbagd import RatbagErrorCode, RatbagdDevice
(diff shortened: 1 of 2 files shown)


Hjdskes:
After merging this change in the bindings needs to be synced to libratbag.

Hjdskes:
Question: do we still want to show the notification when a commit fails? We could capture the `Resync` signal and take this as an indication of failure. Otherwise, I'll need to add a followup commit that removes this notification widget and its logic from the mouse perspective.

Hjdskes:
@whot ping.

whot:
Yeah, we should definitely hook up `Resync` to the notification now. 

Sorry about the delay, things are a tad busy here, merged now.

Hjdskes:
> Sorry about the delay, things are a tad busy here, merged now.

No worries, just wanted to make sure you've seen this and it didn't slip through.

Issue #267, opened 2018-10-05
Remove our appmenu


Hjdskes:
See https://gitlab.gnome.org/GNOME/Initiatives/issues/4

Wiki page: https://gitlab.gnome.org/GNOME/Initiatives/wikis/App-Menu-Retirement

Our TODO:
- [x] Add a Hamburger Menu that contains the current AppMenu items
- [x] Drop the AppMenu

Hjdskes:
This is now tracked in #285, closing it here.

PR #285, opened 2019-01-09
Remove appmenu in line with GNOME developments

Hjdskes:
There's an initiave in GNOME to replace app menus with "in-window" menus: https://gitlab.gnome.org/GNOME/Initiatives/issues/4. This PR brings Piper in line with this new approach, see also #267.

In short, this PR removes our appmenu and adds a menu button to the right-most position in all headerbars (i.e., in the error perspective, the welcome perspective, and the mouse perspective). Right now, all these menus are identical so this works without much hassle (I don't envision us adding any perspective-specific items in the foreseeable future, either).

The menu is a bit.. empty. Since the new guidelines recommend removing the Quit item, we now only have an "About Piper" (not just "About", per the new guidelines) item.


piper/window.py
```diff
@@ -183,3 +173,16 @@ def _on_device_selected(self, perspective, device):
...
+
+    def _perspective_add_back_button(self, perspective):
+        button_back = Gtk.Button.new_from_icon_name("go-previous-symbolic",
+                                                    Gtk.IconSize.BUTTON)
+        button_back.set_visible(len(self.ratbag.devices) > 1)
+        button_back.connect("clicked", lambda button, ratbag:
+                            self._present_welcome_perspective(ratbag.devices),
+                            self.ratbag)
+        self.ratbag.connect("notify::devices", lambda ratbag, pspec:
+                            button_back.set_visible(len(ratbag.devices) > 1))
```
whot: there is no `self.ratbag`, we only connect the signals to it and then drop it. You'll have to pass this down as argument to the function. And please test it, I can't see how this would've worked, maybe you forgot to commit some local change?


Coincidentally, can you in the process fix the indentation so it reads
```
button_back.connect("clicked", 
                    lambda button, ratbag: self._present....,
                    ratbag)
```
Because right now it looks like the `self._present` line is an argument to `connect()`

Hjdskes: > And please test it, I can't see how this would've worked, maybe you forgot to commit some local change?

Sorry, I tested the back button by making it visible manually with GTK's inspector, instead of spawning the ratbagd test device (i.e., I was being lazy).

I fixed this and the indentation.

Hjdskes:
@whot ping

whot:
100% less crashes, I like it! :)

PR #304, opened 2019-02-20
Update for libratbag removing device capabilities

Hjdskes:
This PR includes the changes we need to make in order to work with [libratbag removing device capabilities](https://github.com/libratbag/libratbag/pull/631). Note that I didn't include the changes to ratbagd.py in this PR, nor [the changes I had to make to it in order to show profile names](https://github.com/libratbag/libratbag/pull/631#issuecomment-465549800).

It'd be good if someone could test this on a device that doesn't allow one or more of the following:

1. resolution changes
2. button remapping
3. led changes

My device supports all three and it works on here.

### DIFF
--- a/piper/mouseperspective.py
+++ b/piper/mouseperspective.py
@@ -115,7 +115,7 @@ class MousePerspective(Gtk.Overlay):
                 self.listbox_profiles.select_row(row)
 
     def _hide_notification_error(self):
-        if self._notification_error_timeout_id is not 0:
+        if self._notification_error_timeout_id != 0:
             GLib.Source.remove(self._notification_error_timeout_id)
             self._notification_error_timeout_id = 0
         self.notification_error.set_reveal_child(False)
--- a/piper/mouseperspective.py
+++ b/piper/mouseperspective.py
@@ -19,7 +19,6 @@ from gettext import gettext as _
 from .buttonspage import ButtonsPage
 from .gi_composites import GtkTemplate
 from .profilerow import ProfileRow
-from .ratbagd import RatbagdDevice
 from .resolutionspage import ResolutionsPage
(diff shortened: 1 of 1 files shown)


piper/mouseperspective.py
```diff
@@ -83,20 +82,18 @@ def device(self):
...
         self.stack.foreach(Gtk.Widget.destroy)
-        if RatbagdDevice.CAP_RESOLUTION in capabilities:
+        active_profile = device.active_profile
+        if len(active_profile.resolutions) > 0:
             self.stack.add_titled(ResolutionsPage(device), "resolutions", _("Resolutions"))
-        if RatbagdDevice.CAP_BUTTON in capabilities:
+        if len(active_profile.buttons) > 0:
             self.stack.add_titled(ButtonsPage(device), "buttons", _("Buttons"))
-        if RatbagdDevice.CAP_LED in capabilities:
+        if len(active_profile.leds) > 0:
```
whot: fwiw, the pythonic way here is a simple `if active_profile.leds:` because an empty list is false. The main reason this is superior is efficiency, `len()` has to count the elements only for the result to be thrown away whereas the other option immediately returns when one element is found.

Hjdskes: Thanks, dabbling too much with other languages atm :)

whot:
fwiw, I can't merge this, it still looks like a draft PR

Hjdskes:
That's because https://github.com/libratbag/libratbag/pull/631 isn't in yet, but I'll undraft this if you promise to behave :)

whot:
Merged, thanks. fwiw, doesn't matter that the PR wasn't merged there yet, this will work either way.

PR #305, opened 2019-02-21
Add SVG handling to piper

whot:
This is the other half to the libratbag commit: https://github.com/libratbag/libratbag/pull/638

It adds the SVGs from libratbag and in the second commit hooks up the SVG loading through gresources.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -13,13 +13,6 @@ about_dialog = configure_file(input: 'ui/AboutDialog.ui.in',
 			      output: 'AboutDialog.ui',
 			      configuration: conf)
 
-gnome.compile_resources('piper', 'piper.gresource.xml',
-			source_dir: '.',
-			dependencies: [about_dialog],
-			gresource_bundle: true,
-			install: true,
-			install_dir: pkgdatadir)
-
 install_data('org.freedesktop.Piper.svg', install_dir: icondir)
 
 i18n.merge_file(input: 'org.freedesktop.Piper.desktop.in',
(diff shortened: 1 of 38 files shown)


data/meson.build
```diff
@@ -39,3 +32,60 @@ i18n.merge_file(input: appdata,
...
+svg_mapping = files('svgs/svg-lookup.ini')
+
+test_svg_files = find_program('svgs/check-svg.py')
+test('check-svg', test_svg_files,
+     args: svgs,
+     env: ['BASEDIR=@0@'.format(meson.current_source_dir())])
+
+gresource = configure_file(input: 'piper.gresource.xml.in',
+                           output: 'piper.gresource.xml',
+                           command: ['generate-piper-gresource.xml.py',
```
Hjdskes: Isn't there a built-in in Meson to do something like this? I haven't checked, just wondering out loud.

whot: I couldn't find one. Replacing one line is easy  enough with `configuration_data` but in this case it's beyond meson's intentionally simple capabillities.

whot:
FTR, @bentiss really dislikes shoving the SVGs into gresources because it requires designers to run meson/ninja to test SVGs. It should be possible to just drop the SVG in place and edit the `.ini` file and off we go.

IMO that can be a follow-up commit to this one, I didn't get to this today.

Hjdskes:
> IMO that can be a follow-up commit to this one, I didn't get to this today.

Agreed, this works for now.

PR #306, opened 2019-02-27
Sync with the new API

whot:
This syncs piper with the new ratbagd API. Not that much to do after all, most of it is hidden away in the bindings anyway.

I think it'll be best to merge this with a merge commit btw, otherwise we have commits that don't work on master.

~Also, setting a button macro failed here,  but I think that's a libratbag issue, still investigating.~ *edit:  Multi-key macros fail, that's a HID++ 2.0 implementation limit in libratbag*

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -23,6 +23,7 @@
 
 import os
 import sys
+import hashlib
 
 from enum import IntEnum
 from evdev import ecodes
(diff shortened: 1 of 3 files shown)


Hjdskes:
> I think it'll be best to merge this with a merge commit btw, otherwise we have commits that don't work on master.

Oops, I acted too fast :(

PR #311, opened 2019-02-28
meson: update minimum requirement

Hjdskes:
Same as https://github.com/libratbag/libratbag/pull/663


whot:
let's copy the version check from libratbag to piper, I think that's the better approach here.

Hjdskes:
We have this, so closing.

PR #314, opened 2019-03-02
data: symlink logitech-g900-wireless to logitech-lightspeed

FFY00:
Following https://github.com/libratbag/libratbag/pull/660

### DIFF
--- a/data/svgs/svg-lookup.ini
+++ b/data/svgs/svg-lookup.ini
@@ -70,7 +70,7 @@ Svg=logitech-g703.svg
 DeviceMatch=usb:046d:c081
 Svg=logitech-g900.svg
 
-[Logitech G900, G903, G603, G403]
+[Logitech Device (via Lightspeed connector)]
 DeviceMatch=usb:046d:c539
 Svg=logitech-g900.svg
 


data/meson.build
```diff
@@ -54,6 +54,7 @@ svgs = [
 	'svgs/logitech-g9.svg',
 	'svgs/logitech-g-pro.svg',
 	'svgs/logitech-g-pro-wireless.svg',
+	'svgs/logitech-lightspeed.svg',
```
whot: That's not enough anymore now that the SVGs must be handled  by piper. The lookup table is in `data/svgs/svg-lookup.ini` and the only info we get is `usb:046d:c539` from ratbagd. So we'll look up that string in the .ini file and come up with the `logitech-g900.svg` file. So that symlink doesn't do anything bug exercise the file system on install a bit. Which is outside of piper's scope ;)

You'll need to update the .ini file together with any file changes.

FFY00:
This is all right? Just needed to rename the Device for proper identification? This seems a bit duplicated between piper and libratbag.

data/svgs/svg-lookup.ini
```diff
@@ -70,7 +70,7 @@ Svg=logitech-g703.svg
 DeviceMatch=usb:046d:c081
 Svg=logitech-g900.svg
 
-[Logitech G900, G903, G603, G403]
+[Logitech Device (via Lightspeed connector)]
 DeviceMatch=usb:046d:c539
 Svg=logitech-g900.svg
```
whot: The `Svg` entry still points to the `logitech-g900.svg` file and the new one isn't used anywhere afaict from this diff.

It helps if you forget anything about what libratbag does and focus only on piper. The only identifier we get from ratbagd is the `usb:1235:abcd` string. Now we need to figure out which SVG to render for that string. And that's the `Svg=` entry. The .ini file we have here uses the section headers so that the puny humans can figure out what's going on (instead of say a comment). The rest is just a lookup between the match string and whatever svg we want to render. Changing the section header is like changing a comment but has no functional implications.

whot:
> This seems a bit duplicated between piper and libratbag.

yeah, that's one of the disadvantages of moving it over. Though the piper bits use the device names as section headers which are otherwise unused, so they simply serve to make it easier for puny humans.

FFY00:
So this PR isn't needed at all. Why are we passing the usb id? The same usb id can be used by different mice. Shouldn't we be passing an internal device name instead? How will piper handle the device identification when it gets merged to the kernel driver? Also, a receiver can have several mice connected, passing the usb id to piper doesn't address this.

whot:
once the kernel supports it, the devices will have their own id, just like the current unifying receivers and we can match on that. for the devices where we the vendor really re-uses ids we have a generic version number too, it's just currently unused in piper (see https://github.com/libratbag/libratbag/pull/638)

>  Shouldn't we be passing an internal device name instead?

it's unreliable because the 2018 model of the G123 may be different to the 2019 model of the G123 and then you cannot figure out which one to add to. see https://github.com/libratbag/libratbag/pull/638 for more explanations.

> So this PR isn't needed at all. 

well, right now we load the g900.svg for the lightspeed devices. I still think we should rename it to `...-lightspeed.svg` and fix all occurences in the .ini to use that directly, so nothing uses g900.svg. Once we have the split in the kernel, we can add the devices again with the right USB id.

PR #315, opened 2019-03-03
svgs: use /usr/bin/env shebang for python3 scripts

dguglielmi:


dguglielmi:
Because /bin/python3 is not provided on somes distros (ex: Gentoo), this PR will fix building Piper on that kind of distros.

PR #320, opened 2019-03-05
data: add Logitech G100s

FFY00:
I noticed the Logitech G Pro's SVG could be repurposed for this. What do you think?

### DIFF
--- a/data/svgs/svg-lookup.ini
+++ b/data/svgs/svg-lookup.ini
@@ -2,6 +2,10 @@
 DeviceMatch=usb:1ea7:4011
 Svg=etekcity.svg
 
+[Logitech G100s]
+DeviceMatch=usb:046d:c247
+Svg=logitech-g-pro.svg
+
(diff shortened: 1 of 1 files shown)


whot:
Google suggests the G100s doesn't have the thumb buttons and has no LEDs. So we could re-use the base image and remove those which should only take a few min. But re-using it directly is not ideal, sorry.

Also, the G100s got reverted because it uses a different protocol, should probably wait until libratbag supports again.

FFY00:
Yes, I agree we should wait. Are you talking about the lines? Can't those be hidden in software?

whot:
the lines and the buttons themselves as well. and yes, they could be hidden in software but it takes about 5 min to remove those in inkscape and never think about it again. it takes more than 5 min to write the software to hide those lines on demand, etc. :)

FFY00:
Oh, forgot about the buttons. Makes sense. I am closing this as I don't expect the G100s to be supported that soon.

PR #324, opened 2019-03-19
Fake a profile name where none exists

### LINKED ISSUE #323: Fallback for Empty Profile Names
Is there any way to code the profile switcher menu so that, in the event of the profile names being unset, it just gives the profile number? Like, "Profile 0", "Profile 1", "Profile 2" in the case of my Logitech G502. Currently libratbag can't read the profile names for the G502 so when you request the profile name it comes back as an empty string. As a result it's hard to tell there's even a profile button there, and it's difficult to tell what profile you're currently modifying. It'd be nice if there was a fallback behavior when there's no provided profile name.

Examples:
Screenshot from 2019-03-18 04-38-50
Screenshot from 2019-03-18 04-38-57

whot:
Where the profile name is none (the empty string), fake a profile name based
on the index. This used to be done in ratbagd but was removed there with the
most recent snapshot.

Fixes #323

### DIFF
--- a/piper/mouseperspective.py
+++ b/piper/mouseperspective.py
@@ -94,7 +94,10 @@ class MousePerspective(Gtk.Overlay):
             self.stack.add_titled(LedsPage(device), "leds", _("LEDs"))
 
         self.button_profile.set_visible(len(device.profiles) > 1)
-        self.label_profile.set_label(active_profile.name)
+        name = active_profile.name
+        if not name:
+            name = 'Profile {}'.format(active_profile.index)
+        self.label_profile.set_label(name)
         self._on_profile_notify_dirty(active_profile, None)
 
         # Find the first profile that is enabled. If there is none, disable the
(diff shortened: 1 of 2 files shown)


whot:
IMO not if we want to keep synching it with libratbag. the whole "no name means not writable" is part of the dbus api now, and that's basically what ratbagd.py is for - wrapping that API.

Hjdskes:
I'd argue that, in Piper's case at least, ratbagd.py also exists to wrap the API in a convenient manner. I think this qualifies as such a case.

whot:
Yeah, I understand the argument but this is about keeping `ratbagd.py` useful for *both* libratbag and piper. And the next upcoming change (once I find time to implement it) is to make properties' existence conditional. So rather than an empty string, we'll end up with no `.name` property and that's the signal it doesn't exist. That part will definitely have to go into piper, so I don't think it's useful to have an interim solution pushed to `ratbagd.py`.

Hjdskes:
Fair enough, let's get this in :)

PR #326, opened 2019-03-20
SVG file for Roccat Kone Pure.

hipnoizz:
I'm working on adding support for Roccat Kone Pure in libratbag. In the meantime I'm sending SVG for the mouse to get some feedback.

Questions:

1. I've noticed that current SVGs generally does not have 'buttons' for wheel up/down. Kone Pure allows to reconfigure those buttons and I wonder what would be the best way to show that in SVG. Right now I copied tilt left/right arrow from some other SVG but I'm not that happy with the result. See the attached screenshot.
2. Logo LEDs on other mice are shown in a generic way - usually a circle-like component. Is it due whatever copyright reasons or just so far nobody could be bothered about providing a more 'realistic' logos? (I know that generally the current approach is to have simple, schematic-like images.)

Let me know if you have any other thoughts.

roccat-kone-pure-buttons

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -58,6 +58,7 @@ svgs = [
 	'svgs/logitech-mx-anywhere2.svg',
 	'svgs/logitech-mx-master-2s.svg',
 	'svgs/logitech-mx-master.svg',
+	'svgs/roccat-kone-pure.svg',
 	'svgs/roccat-kone-xtd.svg',
 	'svgs/steelseries-kinzu-v2.svg',
 	'svgs/steelseries-rival310.svg',
--- a/data/svgs/roccat-kone-pure.svg
+++ b/data/svgs/roccat-kone-pure.svg
@@ -1,0 +1,445 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   version="1.1"
+   id="svg64"
+   width="500"
+   height="500"
+   viewBox="0 0 500 500"
+   sodipodi:docname="roccat-kone-pure.svg.2019_03_20_20_26_32.0.svg"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14"
+   style="enable-background:new">
+  <metadata
+     id="metadata70">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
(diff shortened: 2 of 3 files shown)


Hjdskes:
Thanks for doing this! The answer to your second question is that yes, this is for copyright reasons. Technically, we'd be copying, for example, Logitech's logo and attaching that to our product. They might not like that, and that could get us into some issues that we'd rather avoid. So, while I like what you did with the logo (it looks really good!), please switch it to something more abstract. It doesn't necessarily need to be the same as on the other devices, if you want to stick to the design of the mouse.

As for your first question, I believe that ratbag made a conscious desicion to not remap the scroll buttons. @whot please confirm this or tell me to shut it :)

hipnoizz:
Thx for the reply.

RE the logo LED: I will replace it with generic round-like stuff, no problem.

RE disallow mapping wheel up/down buttons: I haven't looked at the core ratbag code but when I put the mapping in the `roccat` driver and placed the buttons in SVG I can remap tem from Piper. Anyway - even if I can find two a bit uncommon cases where would be nice to do that - I could agree that disabling remapping those buttons makes sense.

(Those 2 cases being: a) when the buttons was remapped, e.g. using a software from the vendor and then a user wants to 'fix' that' and b) when playing some old (really old) games when the mouse wheel isn't supported so you can remap wheels to some key shortcuts. For the first scenario I plan to ask ratbag guys if they are interested in adding support for device reset - at least some Roccat devices support that. For the second scenario - let's say it pretty niche.)

Anyway, let me know what is the consensus around mapping the scroll buttons.

whot:
> As for your first question, I believe that ratbag made a conscious desicion to not remap the scroll buttons. @whot please confirm this or tell me to shut it :)

I honestly don't remember. @bentiss?  Maybe this was a decision we made for piper? I'm not sure ratbag has any code specifically to prevent it, but I'm also pretty sure it has no code to specifically enable it. Either way, not remapping wheels makes sense to me because it's more trouble than it's worth IMO.


Roccat uses a lion, right? Just for laughs and giggles, you could add an abstract logo similar to [this one I found on google](https://i-h2.pinimg.com/236x/5f/7b/12/5f7b127b254f3b334fcec67db0137c76--castle-logo-clever-logo.jpg). It's almost the same, right? ;) And I'm only half-joking here.

bentiss:
> I honestly don't remember. @bentiss? Maybe this was a decision we made for piper? I'm not sure ratbag has any code specifically to prevent it, but I'm also pretty sure it has no code to specifically enable it. Either way, not remapping wheels makes sense to me because it's more trouble than it's worth IMO.

I remembered something that it was a little bit of both: some drivers do not allow to remap the wheel up and down, and IIRC we kind of removed their mapping to have something consistent and remove a little bit of extra leaders and buttons to map.

But either way, there is nothing that prevents from libratbag to remap these functions, besides the mouse itself (well, the protocol used by the mouse). So it is a design choice from piper to make.
(in the same way, with `ratbagctl` you can remove the click left from a mouse, making it harder to use, piper disallows this)

gh7531:
What is the status of this PR? 
I'm having a lot of issues with the Roccat Kone software for Linux and was hoping to use piper instead

stephanlachnit:
> What is the status of this PR?
> I'm having a lot of issues with the Roccat Kone software for Linux and was hoping to use piper instead

Right now there isn't even support for the Kone Pure in libratbag. If you want to see support for it, maybe you can help. Just build libratbag with this PR https://github.com/libratbag/libratbag/pull/861 and piper with this PR https://github.com/libratbag/piper/pull/439, test it and leave feedback.

FFY00:
I am closing this in favor of #439.

PR #328, opened 2019-03-23
Update Piper Flatpak to GNOME 3.32

Hjdskes:
See https://github.com/flathub/org.freedesktop.Piper/pull/4


whot:
not sure how I missed that, sorry. thanks for the PR

PR #371, opened 2019-07-23
Add missing device id for MX Master

ngseer:
Update `DeviceMatch` value for Logitech MX Master in according [with the libratbag device description](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-MX-Master.device).

It was first mentioned in [this issue](https://github.com/libratbag/libratbag/issues/439) and I also can confirm that my MX master has `bluetooth:046d:b017` id.

### DIFF
--- a/data/svgs/svg-lookup.ini
+++ b/data/svgs/svg-lookup.ini
@@ -111,7 +111,7 @@ DeviceMatch=bluetooth:046d:b019;usb:046d:4069
 Svg=logitech-mx-master-2s.svg
 
 [Logitech MX Master]
-DeviceMatch=usb:046d:4041;bluetooth:046d:b012;usb:046d:4060
+DeviceMatch=usb:046d:4041;bluetooth:046d:b012;usb:046d:4060;bluetooth:046d:b017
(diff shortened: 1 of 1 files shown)


ngseer:
[Thoughts and possible discussion starter] Still, having two devices-to-ids lists without any guaranteed consistency seems a little inconvenient to me. There might be some nice way to keep these lists synchronized. 

I think I can work on some PR implementing that but we definitely need to come up with some solution first.

whot:
The basic issue is: those lists are from two different projects and have two different purposes.

In libratbag, the list is "these are the devices that have these particular driver settings". In piper, the list is "these are the devices that use this particular SVG image". Except for the general format, there's nothing that's really identical about this.

e.g. libratbag may have to use different device files for two revisions of the same mouse if the protocol changes. Piper can use the same svg. Likewise, libratbag may use the same device file for multiple different mice that all behave the same. Piper would have to use different entries since the SVGs would look different.

So the only long-term meaningful check would be: is there a device ID in libratbag that is missing in Piper. That's the bit we would want to pick up.

PR #381, opened 2019-08-12
circleci: move existing config to the new location

whot:
CircleCI 2.0 requires a new config file location, otherwise it won't run. And
since 1.0 is sunset now, this means our CI wasn't getting executed.

The actual format of the config is the same, we were already using 2.0
features.


bentiss:
LGTM

PR #403, opened 2019-10-04
Add Logitech MX Master 3 SVG

aymanbagabas:
MX Master 3 and 2s have the same button layout but slightly different shape. Use 2s svg for now.


whot:
CI fails because the SVG isn't in meson.build, please add it to that list too (`ninja -C builddir test` will check all these little nitpicks). Thanks.

aymanbagabas:
> CI fails because the SVG isn't in meson.build, please add it to that list too (`ninja -C builddir test` will check all these little nitpicks). Thanks.

Got it! Fixed, rebased, and pushed. Thank you!

PR #420, opened 2019-11-04
G602 now uses the G604 svg

dhollinger:
After the amazing work that @markusressel did with the G604, I took that file and made some slight modifications to create an svg for the G602.

I would venture to guess that this PR is dependent on https://github.com/libratbag/libratbag/pull/838

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -51,6 +51,7 @@ svgs = [
 	'svgs/logitech-g500s.svg',
 	'svgs/logitech-g502.svg',
 	'svgs/logitech-g600.svg',
+	'svgs/logitech-g602.svg',
 	'svgs/logitech-g603.svg',
 	'svgs/logitech-g604.svg',
 	'svgs/logitech-g700.svg',
--- a/data/svgs/logitech-g602.svg
+++ b/data/svgs/logitech-g602.svg
@@ -1,0 +1,753 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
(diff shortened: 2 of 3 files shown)


markusressel:
Is the SVG file any different than the one from the G604? Because if it is not I think they use symlinks instead of copying the file.

dhollinger:
@markusressel It is. I was going to symlink it, but the G602 swaps G7 with G9, G4 with G6, and the `+` and `-` buttons are G10 and G11 respectively

data/meson.build
```diff
@@ -51,6 +51,7 @@ svgs = [
 	'svgs/logitech-g500s.svg',
 	'svgs/logitech-g502.svg',
 	'svgs/logitech-g600.svg',
+    'svgs/logitech-g602.svg',
```
whot: check the indentation on this one please

data/svgs/README.md
```diff
@@ -82,6 +82,7 @@ as of the time of import. See the libratbag COPYING file for details.
  logitech-g500s.svg
  logitech-g502.svg
  logitech-g600.svg
+ logitech-g602.svg
```
whot: this list doesn't need to be updated, it's just the list of devices imported from libratbag which doesn't apply here

whot:
> I would venture to guess that this PR is dependent on libratbag/libratbag#838

fwiw that is true for device support but they're independent otherwise, i.e. we can merge this into piper without having to wait for libratbag.

dhollinger:
@whot It should be good to go now. After that PR was closed with the explanation, I updated the device id in this PR to match what's in libratbagd.

whot:
merged, thanks!

PR #421, opened 2019-11-04
meson.build: use a custom message for a missing python3 dependency

whot:
It's confusing to users because our dependency is the python3 development
package, not the normal python3 package. Let's use a custom message here to
avoid further bug reports about this.


whot:
obsolete with #424

PR #424, opened 2019-11-10
python development package not necessary

stephanlachnit:
Package builds fine without development dependencies since the build process only runs ordinary python scripts. This would also make #421 unnecessary.


whot:
huh, yeah. I wonder why this was always in there, maybe some leftover from the templates generation. thanks!

PR #426, opened 2019-11-16
Fix license changes for svgs

stephanlachnit:
piper is licensed under GPLv2, not under GPLv3.

### DIFF
--- a/data/svgs/README.md
+++ b/data/svgs/README.md
@@ -101,5 +101,5 @@ as of the time of import. See the libratbag COPYING file for details.
  steelseries-senseiraw.svg
 
 Where changes are made to these files, the changes will be made under
-piper's license (GPLv3). Use the git commit log to determine the exact
+piper's license (GPLv2). Use the git commit log to determine the exact
 license state of each file.


stephanlachnit:
Since I'm trying to maintain an official Debian package: would it be possible (if legally allowed) to re-license all svgs under GPLv2? That would make it way easier to maintain the package, since I wouldn't have to check if one of the original svgs has been changed for Debian's copyright file.

whot:
thanks! 

as for the other question - As I understand, MIT is compatible with GPLv2 so while we must not remove the MIT license itself, we can add the GPLv2 and be done with that bit. It would be useful to contact the various authors to ask for permission to remove the MIT for no other reason than to remove clutter. I'm personally fine with that for the ones I did though we'd have to go through the list of changes before the import to catch all authors.

> See the libratbag COPYING file for details.

libratbag's file could theoretically change and for a legal document we shouldn't point to something outside of our repo. Let's just add that file as `COPYING.MIT` to the repo, one less thing to think about.

stephanlachnit:
> as for the other question - As I understand, MIT is compatible with GPLv2 so while we must not remove the MIT license itself, we can add the GPLv2 and be done with that bit. It would be useful to contact the various authors to ask for permission to remove the MIT for no other reason than to remove clutter. I'm personally fine with that for the ones I did though we'd have to go through the list of changes before the import to catch all authors.
> 
> > See the libratbag COPYING file for details.
> 
> libratbag's file could theoretically change and for a legal document we shouldn't point to something outside of our repo. Let's just add that file as `COPYING.MIT` to the repo, one less thing to think about.

Perfect, is the wording I proposed okay? Then I would open a PR with the change and the added MIT license. When it comes to the removal of the MIT license: since I'm not a maintainer of this software I prefer not get involved, although I can help if wanted.

whot:
yep, wording sounds fine to me, please file a PR and cc me on it. I'll see about the MIT removal but it'll take a while, not the top-most priority right now :)

stephanlachnit:
> I'll see about the MIT removal but it'll take a while, not the top-most priority right now :)
It's not that important to me either. As long as they can also be licensed under GPLv2 I don't have a messy copyright file in Debian.

Btw the Debian package got released: https://tracker.debian.org/pkg/piper, and it is also included in the Ubuntu Focal Fossa universe repository: https://launchpad.net/ubuntu/+source/piper
That's great news for piper, and the installation wiki page can be updated correspondingly!

whot:
done, thanks!

PR #441, opened 2019-12-09
Make builds reproduceable

stephanlachnit:
The Debian package tracker [noticed](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/piper.html) that the builds are not reproduceable, meaning they can change under certain scenarios. I checked it and it is the copyright year in the AboutDialog, which was set to change dynamically in f2a9e5d8bbf30360ea207a4859573ef2f1b0470f. While it sounds like a good idea to not put it in the file, an old version shouldn't have a copyright note which is never than the actual release date in my opinion. I put the `copyright_years` variable below the `version_date` variable so it won't get forgotten in new releases.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -4,11 +4,8 @@ desktopdir = join_paths(datadir, 'applications')
 icondir = join_paths(datadir, 'icons', 'hicolor', 'scalable', 'apps')
 metainfodir = join_paths(datadir, 'metainfo')
 
-date = run_command('date', '+%Y')
-
 conf = configuration_data()
 conf.set('version', meson.project_version())
-conf.set('year', date.stdout().strip())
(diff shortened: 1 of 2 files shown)


whot:
ACK but I really wonder whether it's just simpler to leave this out. The whole `Copyright 20whenver` is borderline useless anyway when it's set automatically and as generically as we do here.

How does it look like when we remove that line altogether? Does the dialog box still work fine?

FFY00:
Sure, I think that is fine.

stephanlachnit:
> How does it look like when we remove that line altogether? Does the dialog box still work fine?
Tested it and it works fine.

whot:
Thanks, merged!

PR #443, opened 2019-12-10
Update CircleCI config

stephanlachnit:
I looked into the CircleCI config and found some room for improvements:

1. It's still using Fedora 30, latest is Fedora 31.
2. The caching workflow doesn't really make sense because we don't have multiple Fedora instances, thus it simply increases the time (probably just imported from libratbag, where that's the case)
3. Since Fedora Rawhide isn't used anyway the config can be removed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,7 @@ piper_references:
     working_directory: ~/piper
     environment:
       LANG: C.UTF-8
+
   build_default: &build_default
     name: Build
     command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,5 +1,3 @@
-# vim: set expandtab shiftwidth=2 tabstop=8:
-
```
whot: why did you remove this one? it's really handy to have the editor do the right thing when opening the file

stephanlachnit: I removed it because it wasn't mentioned in the commit in libratbag that added it and thought it was maybe unintended (also I don't use vim), but I can restore it.

whot: it's a useful thing to have. There's a line for emacs too to auto-configure special settings for the opened files but neither @bentiss nor I use emacs, so .. meh :) And these days' there's also [editorconfig](https://editorconfig.org/#example-file) but that's less useful when there's a single file with different requirements.

whot:
merged, thanks. fwiw, I added your MR comments to the commit message, that way they're preserved for posterity :)

PR #448, opened 2019-12-12
ResolutionRow Redesign

stephanlachnit:
This PR does two main things:
- Redesign the ResolutionRow glade file to support disabling a resolution and setting a y-resolution
- Implement the DBus support for this

Requires libratbag/libratbag#870 to be merged.

### DIFF
--- a/data/ui/ResolutionRow.ui
+++ b/data/ui/ResolutionRow.ui
@@ -15,14 +15,13 @@
             <property name="visible">True</property>
             <property name="can_focus">False</property>
             <property name="border_width">12</property>
+            <property name="spacing">6</property>
             <child>
               <object class="GtkLabel" id="dpi_label">
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">start</property>
-                <property name="label">2800 DPI</property>
-                <property name="single_line_mode">True</property>
-                <property name="track_visited_links">False</property>
+                <property name="label">0 DPI</property>
               </object>
               <packing>
                 <property name="expand">True</property>
@@ -32,11 +31,11 @@
             </child>
             <child>
               <object class="GtkLabel" id="active_label">
+                <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">end</property>
-                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active resolution.">Active</property>
+                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active resolution.">active</property>
                 <property name="justify">right</property>
-                <property name="track_visited_links">False</property>
               </object>
               <packing>
                 <property name="expand">False</property>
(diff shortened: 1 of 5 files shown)


stephanlachnit:
- Doesn't work in it's current state, sets dpi to lowest possible value instead of 0
- Active profile indication needs redesign, button is probably better.

stephanlachnit:
So here is my idea for the new UI:
new_ui

I added another scale so we can set separate xy resolutions. The buttons are toggle buttons. Feedback is appreciated.

stephanlachnit:
Wohoo the UI is working now! It's actually already usable, the only thing is that a disabled resolution will be set as the lowest possible dpi. The only implementation left is the checks for the capabilities. 
The resolutionpage should re-initialize itself after the values are applied. The goal here is that if a driver actually removes a dpi instead of disabling it, piper should remove the row. However this also requires an option to add new resolutions, and I won't add this for now.

data/ui/ResolutionRow.ui
```diff
@@ -31,12 +30,12 @@
               </packing>
             </child>
             <child>
-              <object class="GtkLabel" id="active_label">
+              <object class="GtkLabel" id="status_label">
+                <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">end</property>
-                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active resolution.">Active</property>
+                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active status.">status</property>
```
stephanlachnit: Will break translations. Probably not fixable since the label text changes dynamically. Any ideas?

piper/ratbagd.py
```diff
@@ -428,7 +428,7 @@ def capabilities(self):
         """The capabilities of this profile as an array. Capabilities not
         present on the profile are not in the list. Thus use e.g.
 
-        if RatbagdProfile.CAP_WRITABLE_NAME is in profile.capabilities:
+        if profile.CAP_WRITABLE_NAME in profile.capabilities:
```
stephanlachnit: Whoops leftover from #450. Will revert in next revision.

piper/ratbagd.py
```diff
@@ -550,6 +553,16 @@ def _on_properties_changed(self, proxy, changed_props, invalidated_props):
...
 
+    @GObject.Property
+    def capabilities(self):
+        """The capabilities of this resolution as an array. Capabilities not
+        present on the resolution are not in the list. Thus use e.g.
+
+        if resolution.CAP_DISABLE in resolution.capabilities:
+            do something
+        """
+        return self._get_dbus_property("Capabilities") or []
```
stephanlachnit: How does this part work? What is the actual DBus call in the end? I'm super confused.

piper/ratbagd.py
```diff
@@ -572,7 +585,7 @@ def resolution(self, resolution):
         @param res The new resolution, as (int, int)
         """
         res = self.resolution
-        if len(res) != len(resolution) or len(res) > 2:
+        if len(res) > 2:
```
stephanlachnit: Length will be checked properly with capability.

piper/resolutionrow.py
```diff
@@ -59,16 +67,31 @@ def _init_values(self, resolution):
...
+        with self.xscale.handler_block(self._xscale_handler):
+            self.xscale.props.adjustment.configure(xres, minres, maxres, 50, 50, 0)
+            self.xscale.set_value(xres)
+        with self.yscale.handler_block(self._yscale_handler):
+            self.yscale.props.adjustment.configure(yres, minres, maxres, 50, 50, 0)
+            self.yscale.set_value(yres)
+        if xres == 0:
+            self.disable_togglebutton.set_active(True)
+        self.update_resolution_values(xres, yres)
+        self._on_is_active_changed(resolution, pspec=None)
```
stephanlachnit: It works just fine like this, but what is pspec?

piper/resolutionrow.py
```diff
@@ -97,43 +120,92 @@ def _on_change_value(self, scale, scroll, value):
         idx = self.resolutions.index(value)
         if idx < len(self.resolutions) - 1:
             delta = self.resolutions[idx + 1] - self.resolutions[idx]
-            self.scale.props.adjustment.set_step_increment(delta)
-            self.scale.props.adjustment.set_page_increment(delta)
+            scale.props.adjustment.set_step_increment(delta)
```
stephanlachnit: Any difference if self or not?

piper/resolutionrow.py
```diff
@@ -54,21 +62,39 @@ def _init_values(self, resolution):
...
+        with self.xscale.handler_block(self._xscale_handler):
+            self.xscale.props.adjustment.configure(xres, minres, maxres, 50, 50, 0)
+            self.xscale.set_value(xres)
+        with self.yscale.handler_block(self._yscale_handler):
+            self.yscale.props.adjustment.configure(yres, minres, maxres, 50, 50, 0)
+            self.yscale.set_value(yres)
+        if xres == 0:  # resolution.is_disabled
+            self.disable_togglebutton.set_active(True)
+        self._on_dpi_values_changed(xres=xres, yres=yres)
+        self._on_status_changed(resolution, pspec=None)
```
stephanlachnit: It works just fine like this, but what is pspec?

stephanlachnit:
So I hope the UI is now finally done. Still need to implement the disabled property, but that shouldn't be too hard. UI can already be tested. Would be nice if someone can test it with a mouse with separate xy resolutions. Disabling won't work properly because that requires the property, so keep this in mind when the UI doesn't update correctly here.

stephanlachnit:
Think I'm done :) Is there any way to run `piper.devel` with `ratbagd.devel`?

FFY00:
I am not sure, I tried and it doesn't work. I usually just run the generated `ratbagd`. IIRC the only difference in the `.device` path but I could be wrong.

stephanlachnit:
> I am not sure, I tried and it doesn't work. I usually just run the generated `ratbagd`. IIRC the only difference is the `.device` path but I could be wrong.

But this does require to disable the system installation, right?

FFY00:
No, you can just run it manually `sudo ./ratbagd`. The only thing is that iot will pick up the `.device` files in `/usr/share/libratbag`.

stephanlachnit:
UI doesn't work properly when disabling a resolution. When trying to enable
```
ratbag debug: profile 0 resolution 1: set enabled (libratbag)
ratbag debug: profile 0 resolution 1: set 0 dpi (libratbag)
```

stephanlachnit:
Can be reviewed/merged once libratbag/libratbag#870 is merged.

piper/resolutionrow.py
stephanlachnit: Nevermind, bascially just the trigger.

whot:
>  added another scale so we can set separate xy resolutions. The buttons are toggle buttons. Not supported features (y-scale and disable-button) can be made invisible, we have to check the mouse for features anyway.

one comment here: we *explicitly* didn't support this in piper because many of the mouse features are ... questionable. there may be users for them out there but I have suspicions that some features are merely there so marketing can print another thing on the box that appears to make a device worth buying. Separate x/y resolution is one case like that.

libratbag already cuts a lot of potential features because it'd be a nightmare to support them all. Piper cuts even more because maintaining a useful GUI with too many options is somewhere between hard and ugly (also, ideally *every* feature supported gets tested for every changeset :).

So the difficult choice here to decide which features are worth **not** implementing. Because once you have them you can never remove them without getting abuse (interestingly, the same is not true for features you never supported to begin with). You'll also have to test the combinations - one mouse with the feature, one without that feature. Multiplied by every choice you make somewhere in the UI.
It's a very common mistake to implement features because they're easy to implement (I know, I have made it many times and am now stuck with the effects).

Unless you have a **common** use-case that 90% of your users need or at least want, it's probably not worth implementing. There will always be 10% of potential users that fall outside of what you enable and accommodating for all of them is a neverending timesink.

stephanlachnit:
Sounds reasonable, who uses separate dpis anyway. Will remove it.

stephanlachnit:
Removed y-res support and rebased against latest libratbag PR.

staticssleever668:
Fixed some `Profile.Enabled` instances not being changed to `Profile.Disabled`, would be nice to squash this with the first commit before merging.

staticssleever668:
I just tried this and I really like it. I don't like a few unrelated code changes, but generally it's okay. Hope this gets merged soon.

data/ui/ResolutionRow.ui
```diff
@@ -32,11 +31,11 @@
             </child>
             <child>
               <object class="GtkLabel" id="active_label">
+                <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">end</property>
-                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active resolution.">Active</property>
+                <property name="label" translatable="yes" comments="Translators: this is used to indicate the active resolution.">active</property>
```
staticssleever668, line 37: Note: this would need a translation update

staticssleever668 (approved):

staticssleever668:
Great!

PR #450, opened 2019-12-13
Fix typo in profile.capabilities description

stephanlachnit:
Wrong syntax.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -428,7 +428,7 @@ class RatbagdProfile(_RatbagdDBus):
         """The capabilities of this profile as an array. Capabilities not
         present on the profile are not in the list. Thus use e.g.
 
-        if RatbagdProfile.CAP_WRITABLE_NAME is in profile.capabilities:
+        if profile.CAP_WRITABLE_NAME in profile.capabilities:
             do something
         """
         return self._get_dbus_property("Capabilities") or []


stephanlachnit:
Do not merge. We should edit `ratbagd.py` in libratbag and import it from there. Actually why are we not importing it as a subproject like they did here: https://github.com/FeralInteractive/gamemode/tree/master/subprojects

FFY00:
I don't think this is a typo but it's not my code, so I will leave it to @whot.

stephanlachnit:
> I don't think this is a typo but it's not my code, so I will leave it to @whot.

Pretty sure it is: `is in` is not a valid python call, and it doesn't make sense to differentiate between `RatbagdProfile` (class) and `profile` (variable).

FFY00:
The `is` part might be but the other I am pretty isn't. `CAP_WRITABLE_NAME` is a class variable, not an instance variable. It is present in `profile` because it is instantiated from `RatbagdProfile` but `CAP_WRITABLE_NAME` is not an instance variable, it is not specific for `profile`.

stephanlachnit:
> variable



> The `is` part might be but the other I am pretty isn't. `CAP_WRITABLE_NAME` is a class variable, not an instance variable. It is present in `profile` because it is instantiated from `RatbagdProfile` but `CAP_WRITABLE_NAME` is not an instance variable, it is not specific for `profile`.

That's true, but I thought that this is more an actual code example, and here `RatbagdProfile.CAP_WRITABLE_NAME` will not work afaik.

FFY00:
Why wouldn't it work?

stephanlachnit:
> Why wouldn't it work?

Code returns with `NameError: name 'RatbagProfile' is not defined`. Python works here different than C.

FFY00:
It needs to be imported.

stephanlachnit:
> It needs to be imported.

But that's useless bloat, since you have the variable anyway.

FFY00:
I would not say that. It helps redability, it states where that variable came from.

stephanlachnit:
> I would not say that. It helps redability, it states where that variable came from.

I thought about it and that's actually a very good point. Closing this anyway since it will be included the PRs for the disabled resolutions functionaliy.

whot:
> Do not merge. We should edit ratbagd.py in libratbag and import it from there. 

fwiw, changes to `ratbagd.py` can flow either way.

> Actually why are we not importing it as a subproject [...]

there's still the grand idea on the horizon that once libratbag's API is stable, we don't need to copy `ratbagd.py` back and forth anymore and it can be tailored more precisely to each project.

The `is in` was a typo, yes. The class name was intentional to make it more obvious where it comes from. Since it's documentation, I'm playing a bit loose here with actual code accuracy ;)

PR #451, opened 2019-12-16
AboutDialog: miscm improvements

stephanlachnit:
Adds more readable version string.
More representative website label.
Also fixes translations against commit b02460028568c9f03b752e095264af87450badae.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -6,7 +6,6 @@ metainfodir = join_paths(datadir, 'metainfo')
 
 conf = configuration_data()
 conf.set('version', meson.project_version())
-conf.set('url', 'https://github.com/libratbag/piper')
 conf.set('version_date', version_date)
 
 about_dialog = configure_file(input: 'ui/AboutDialog.ui.in',
--- a/data/ui/AboutDialog.ui.in
+++ b/data/ui/AboutDialog.ui.in
@@ -7,9 +7,9 @@
     <property name="modal">True</property>
     <property name="type_hint">normal</property>
     <property name="program_name">Piper</property>
-    <property name="version">@version@</property>
-    <property name="website">@url@</property>
-    <property name="website_label" translatable="yes">Visit Piper’s website</property>
+    <property name="version" translatable="yes">Version: @version@</property>
+    <property name="website">https://github.com/libratbag/piper</property>
+    <property name="website_label" translatable="yes">Visit Piper on GitHub</property>
     <property name="translator_credits" translatable="yes">translator-credits</property>
     <property name="logo_icon_name">org.freedesktop.Piper</property>
     <property name="license_type">gpl-2-0</property>


po/be.po
```diff
@@ -302,11 +302,11 @@ msgstr "Калі ласка абнавіцеся да апошняй версі
 msgid "Unknown exception occurred"
 msgstr "Адбылася невядомая памылка"
 
-#: data/ui/AboutDialog.ui.in:13
-msgid "Visit Piper’s website"
+#: data/ui/AboutDialog.ui.in:12
+msgid "Visit Piper on GitHub"
```
whot: ok, I freely admit I'm not too well-versed with translations to know this but: isn't this superfluous? and possibly covering up that we need a translation here? Those files are supposed to be re-generated by `ninja` (see the `po/README.md` file) and by updating this bit here, you're effectively hiding that there were changes. Without this commit, the next translation build will show the changed messages in the git diff, making it easier to identify.

But then again, I might be completely wrong about this here :)

stephanlachnit: Oh I haven't checked the README actually, didn't think to much about the translations. Thanks for the heads up, will look into it :+1:

stephanlachnit: Ok so if I get it right, we don't update any translations files, because that's done automatically when running `piper-update-po`, meaning I can simply remove the last commit. On the other hand, are the translations working without the changes?
Is there a way to "force" a specific language? Don't want to change my system language just to test this.
Same problem will occur in #448 btw.

whot:
```
$ export LANG=de_DE.UTF-8
$ piper
```
Doesn't appear work with `piper.devel` run from in-tree unfortunately, the lookup path is wrong. And I can't find whatever meson magic is required to make it work. So you do need to install and run from the prefix dir.

stephanlachnit: Doesn't work for me, stays english :/
But since #448 will require new translations anyway I will just remove the last commit. Is there any release schedule for Piper? Since this commit does nothing super important we could just merge it with #448, so we don't break any translations in a new release.

FFY00: @Hjdskes sorry for pinging you, do you remember how to deal with this?

Hjdskes: With "this", do you mean how to run with in-tree translations?

The lookup path is set to `localedir`, which is [set to `po` for devel builds](https://github.com/libratbag/piper/blob/master/meson.build#L105). I vaguely remember this working for me when I first added translations to Piper. You could see what `localedir` points to in `piper.devel`; it should point to the local `po` folder.

FFY00: Yes, thanks!

stephanlachnit: > Doesn't work for me, stays english :/

Ahhh I don't have any languages installed, that's why it didn't work. Whatever, doesn't matter anyway.

PR #452, opened 2019-12-25
dbus: more gracefully handle permission when connecting to the daemon

theMarix:
This addresses issue libratbag/libratbag/#802 reported for libratbag by porting over the solution implemented in libratbag and adjusting the UI error message to hint at the potential permission issue.

### DIFF
--- a/piper/ratbagd.py
+++ b/piper/ratbagd.py
@@ -252,11 +252,13 @@ class Ratbagd(_RatbagdDBus):
 
     def __init__(self, api_version):
         super().__init__("Manager", None)
-        result = self._get_dbus_property("Devices") or []
-        self._devices = [RatbagdDevice(objpath) for objpath in result]
-        self._proxy.connect("notify::g-name-owner", self._on_name_owner_changed)
+        result = self._get_dbus_property("Devices")
+        if result is None and not self._proxy.get_cached_property_names():
+            raise RatbagdUnavailable("Make sure it is running and your user is in the required groups.")
         if self.api_version != api_version:
             raise RatbagdIncompatible(self.api_version or -1, api_version)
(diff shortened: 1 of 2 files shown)


piper/ratbagd.py
```diff
@@ -28,7 +28,7 @@
 from enum import IntEnum
 from evdev import ecodes
 from gettext import gettext as _
-from gi.repository import Gio, GLib, GObject  # noqa
```
stephanlachnit: Any reason for the change here? If you just want to import the latest `ratbagd.py` from libratbag, then import it completely (line 1 for example) and mention it in the commit description.

theMarix: Reason for this is that I copied over the file and that hunk went under my radar when viewing the commit diff. Thanks for catching it.

theMarix:
Regarding the translation. Should I not modify the message in the error dialog, or should I just not touch the translations right now?

theMarix:
I dropped accidental change to the imports and the translation modifications.

stephanlachnit:
> Regarding the translation. Should I not modify the message in the error dialog, or should I just not touch the translations right now?

Just don't update the translations. I don't know what @whot's plans for a new release are, but we could just ping all translators and ask if they want to update it. #451 requires updated translations as well, so it doesn't really make sense to update them in every PR.

theMarix:
Thanks. I re-added the updated message in the error dialog but skipped updating any translation files.

whot:
> I don't know what @whot's plans for a new release are

I think at this point I'm punting this to @FFY00 because he has vastly more time than me, I barely manage to get the drive-by reviews in.

stephanlachnit:
@whot @bentiss @FFY00 can someone of you merge this? Would make the whole translation effort a little bit easier for the contributors.

PR #460, opened 2020-01-24
po: update German translations

whot:


whot:
I'm completely snowed under right now, if you can fix it please do so, thanks.

Also, fwiw - for the next release it's best to declare a feature-freeze before the translations are merged so we don't need to ask translators to do things multiple times.

stephanlachnit:
Yep the translations were maybe a little bit too rushed from me. Thanks for the invite, I will add the missing lines.

PR #462, opened 2020-01-24
Update Polish translation 200124

piotrdrag:


piotrdrag:
Sure!

po/pl.po
```diff
@@ -529,3 +533,9 @@ msgstr "Wybór urządzenia"
 #: data/ui/Window.ui:8
 msgid "_About Piper"
 msgstr "_O programie"
+
+msgid ""
```
stephanlachnit: Isn't here `#: piper/window.py:55` missing? Anyway, you can rebase against master now since it's merged.

piotrdrag: It doesn’t matter, the .po files will get updated during build anyway.

piotrdrag: OK, rebased!

stephanlachnit: Thanks, didn't know that.

PR #463, opened 2020-01-24
po: Update french translations 

Funeoz:


bentiss:
Thanks. I have manually applied and push this one. It seems there was a merge commit in the middle. So to keep the history clean, I just rebased on top of master and pushed.

stephanlachnit:
Looks like the changes from #452 are missing (should be one or two lines). Can you add them?

PR #466, opened 2020-01-27
Update the Bulgarain translation.

lyubomirv:


stephanlachnit:
Thanks! It's a little bit odd that GitHub shows that the commit changes every line, even though some lines didn't change. But I haven't checked with git yet.

stephanlachnit:
I took your changes in a new commit, the line changes are now fixed. Thanks again!

Issue #478, opened 2020-02-26
Support for Logitech G815


K1ngjulien:
**Is your feature request related to a problem? Please describe.**
I've got a Logitech G815. libratbag detects it since the last update and it also shows up in piper.
Unfortunately, the software thinks the first two macro keys are left and right mouse button, so I can't change them at all and the lighting control doesn't work either.
Furthermore, there's no preview image being loaded.

**Describe the solution you'd like**
I'd like to help add support for this keyboard.
I can start by drawing a image for it in inkscape and then continue with the code but I'd probably need some help with that :)

**Additional context**
![Screenshot_20200226_175523](https://user-images.githubusercontent.com/16562333/75367824-2c192800-58c1-11ea-975e-b20b376e903d.png)

K1ngjulien:
Do you guys even want to support keyboads?

K1ngjulien:
Ok I just figured out that the lighting does work, but I have to press one of the "M-Keys" to refresh the lighting after hitting apply

FFY00:
Yes, unfortunately you need to press one of the M-Keys after saving the changes to the device. This behavior comes from the keyboard, I haven't found any way to get around it.

Your PR looks great! I will test it when I get home.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

agolks:
will it be possible to set colors per key?
currently i need to use g810-led instead of piper to get custom colors.

matrixik:
[...]

PR #479, opened 2020-02-26
data: add SVG for the Logitech G815

K1ngjulien:
adds basic support for the logitech g815 keyboard
https://github.com/libratbag/piper/issues/478

I've created the svg and added the usb device id.

It still thinks its a mouse, therefore it won't let me change G1 and G2 to something other than LMB and RMB.

Weirdly, the macro tab keys (m1-3) are now also normal macro keys, as the macro tab feature is missing and I don't know how to do that.

lastly, I have to press one of the m-keys in order to get the lighting to update after clicking apply.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -52,6 +52,7 @@ svgs = [
 	'svgs/logitech-g604.svg',
 	'svgs/logitech-g700.svg',
 	'svgs/logitech-g703.svg',
+	'svgs/logitech-g815.svg',
 	'svgs/logitech-g900.svg',
 	'svgs/logitech-g-pro.svg',
 	'svgs/logitech-g-pro-wireless.svg',
--- a/data/svgs/logitech-g815.svg
+++ b/data/svgs/logitech-g815.svg
@@ -1,0 +1,1867 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="500"
+   height="500"
+   viewBox="0 0 132.29166 132.29167"
+   version="1.1"
+   id="svg8"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14"
+   sodipodi:docname="logitech-g815.svg">
+  <defs
+     id="defs2" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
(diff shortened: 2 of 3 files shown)


K1ngjulien:
oh and the keyboard rather small in the preview window as its so wide. 

Screenshot_20200226_213213


Screenshot_20200226_213024

Screenshot_20200226_213155

FFY00:
Oh, this actually works better than I was expecting :grin:.

@whot what do you think?

K1ngjulien:
I should probably change the font to a path, will do that this evening.

FFY00:
Just one thing: we don't have the cord plug in any of the devices, would it be possible to remove it from the SVG? This would also allow us to reuse it for the G915.

K1ngjulien:
yeah no problem.

stephanlachnit:
Wow, looks amazing! I thought we had to rewrite how we display devices for keyboards, but I'm not so sure right now.

K1ngjulien:
I guess for keyboards with special macro keys like this one its easy.
Something like the Corsair keyboards, where every key is remappable as a macro key is probably a different story.

stephanlachnit:
> Something like the Corsair keyboards, where every key is remappable as a macro key is probably a different story.

True, but I don't even know if that is something we want to expose, little advantage for most users. We already do this with the main mouse buttons, you can switch between left hand and right hand mode, but you can't unbind them.

FFY00:
Well, Logitech keyboards allow us to remap all keys, but we don't support that.

stephanlachnit:
Btw this would also work for the G413 once https://github.com/libratbag/libratbag/pull/926 is tested and merged.

K1ngjulien:
Would it be possible to change the G1 and G2? I've seen in that the first two buttons are hardcoded to be LMB and RMB and it works if I modify that if.
https://github.com/libratbag/piper/blob/4119897485745912dff604f716d8022fb60f6ff9/piper/buttondialog.py#L112

 there would have to be a way to seperate keyboards from mice, maybe add a Type property in the lookup.ini?
https://github.com/libratbag/piper/blob/master/data/svgs/svg-lookup.ini

stephanlachnit:
Good point. I'd rather have a property in the device file.

K1ngjulien:
Is that the one in libratbag?

stephanlachnit:
Yes, would require changes to libratbag.

K1ngjulien:
anything I can do to help?

stephanlachnit:
If you want, sure. I'm sadly short on time right now, but I can answer your questions if you want to go ahead.
I would say add a .device file entry what "type" the device is, and then adding a property to ratbagd's device definition, so we can check it in Piper.

K1ngjulien:
so in `logitech-g815.device` in libratbag could look like this?:

```
[Device]
Name=Logitech G815
DeviceMatch=usb:046d:c33f
LedTypes=logo;switches
DeviceType=keyboard
Driver=hidpp20
```
Or do I have the completely wrong idea of what I'm supposed to to? ^^

K1ngjulien:
Ok so I just discovered ratbag_device->ratbag_device_data and i'm guessing it needs to be added there as well. I will look into that tomorrow evening.

FFY00:
You need to add it to the config parsing routines (sorry, can't remember names without looking at the code) and export it via DBus.

K1ngjulien:
I've started implementing this. For now I only have the config parsing, as I still have to figure out how DBus works :)

https://github.com/libratbag/libratbag/pull/931

FFY00:
Ah, right. Just reviewed it.

.gitignore
```diff
@@ -0,0 +1 @@
+builddir/
```
stephanlachnit: Use local .gitignores:
```gitignore
.gitignore
builddir/
```

piper/.gitignore
```diff
@@ -0,0 +1 @@
+__pycache__/
```
stephanlachnit: same

PR #494, opened 2020-03-29
Add 250 Hz polling rate

### LINKED ISSUE #491: Set polling rate below 500?
Some games in wine have issues with high polling rates and even 500 will cause some lag, however, this software can only go down to 500 even though in windows it can reach 250

I'd like the ability to change polling rates to 250 and 750 in settings

Currently configuring polling rates in terminal is a chore

stephanlachnit:
Fixes #491. Still need to test it.

### DIFF
--- a/data/ui/ResolutionsPage.ui
+++ b/data/ui/ResolutionsPage.ui
@@ -61,8 +61,8 @@
                         <property name="homogeneous">True</property>
                         <property name="layout_style">expand</property>
                         <child>
-                          <object class="GtkRadioButton" id="rate_500">
-                            <property name="label">500</property>
+                          <object class="GtkRadioButton" id="rate_125">
+                            <property name="label">125</property>
                             <property name="visible">True</property>
                             <property name="can_focus">True</property>
                             <property name="receives_default">False</property>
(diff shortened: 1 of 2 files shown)


stephanlachnit:
I'm not entirely sure why but it doesn't work.

stephanlachnit:
Ok nvm stupid mistake by me, but I get some weird behavior with my G Pro: if I set 250 Hz, ratbagctl reports 200 Hz. @FFY00 any idea why that is the case? @cncz42 can you test it and report as well?

FFY00:
That seems weird.

Can you paste the feature list (easiest way is `solaar show`)?

stephanlachnit:
> That seems weird.

It really doesn't make any sense to me either. Can you reproduce it with any of your devices?

What I just found out, what makes it even weirder: it now default to 250 Hz when reconnecting (I set it once to 250 Hz in Windows), even though I set back to 1000 Hz in Piper.

> Can you paste the feature list (easiest way is `solaar show`)?

`solaar: error: Logitech receiver not found`, I have a G Pro but not the wireless one.

stephanlachnit:
Setting it with this commit to 125 Hz results in `ratbagctl` showing 100 Hz.

staticssleever668:
This is a problem on ratbagd side. https://github.com/libratbag/libratbag/pull/1161 fixes it.

whot:
Fixed with #703, sorry about this lingering so long

PR #496, opened 2020-04-01
data: Add Logitech M720 Triathlon

koefmatt:
Add graphics and device definition for the Logitech M720 Triathlon wireless mouse.

Product page: https://www.logitech.com/en-us/product/m720-triathlon
The DeviceMatch was taken from [libratbag](https://github.com/libratbag/libratbag/blob/master/data/devices/logitech-M720.device)

Bildschirmfoto von 2020-04-01 23-31-02
The button below the wheel is the ratchet mode switch. The "1 2 3" button switches between the three devices the mouse can be connected to.

The mouse has an additional button moulded into the thumb rest which some users report as being mapped to Alt+Tab, however I cannot reproduce this. Remapping this button also doesn't do anything but I think that's an issue with libratbag rather than Piper.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -56,6 +56,7 @@ svgs = [
 	'svgs/logitech-g900.svg',
 	'svgs/logitech-g-pro.svg',
 	'svgs/logitech-g-pro-wireless.svg',
+	'svgs/logitech-m720.svg',
 	'svgs/logitech-mx518.svg',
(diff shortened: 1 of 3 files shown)


FFY00:
> LGTM but can you squash the commit into one?

You can do that when merging, select 'Squash and merge' from the dropdown.

PR #499, opened 2020-04-16
piper.devel: don't write bytecode

stephanlachnit:
Keeps the source dir clean.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -104,6 +104,7 @@ config_piper_devel = config_piper
 config_piper_devel.set('pkgdatadir', join_paths(meson.build_root(), 'data'))
 config_piper_devel.set('localedir', join_paths(meson.build_root(), 'po'))
 config_piper_devel.set('devel', '''
+sys.dont_write_bytecode = True
 sys.path.insert(1, '@0@')
 print('Running from source tree, using local files')
 '''.format(meson.source_root()))


FFY00:
Why? You can do this by using `-B`. There is no reason to explicitly set this.

stephanlachnit:
> Why? You can do this by using `-B`. There is no reason to explicitly set this.

I know, but there is actually is a reason. We simply move the whole `piper` folder when running `ninja install`, so any `__pycache__` artifacts will be installed with that. Ofc you can run `py3clean .` or always start `piper.devel` with `-B`, but it's easy to forget.
As long as there is no disadvantage doing this, I don't see a reason against it.

FFY00:
It does work when changing the code, it will only recompile the changed files. There is a reason `__pycache__` is created by default, and if we have any actual necessity to remove it, I don't think we should.

Actually, it seems like we are not installing `__pycache__` for the normal piper:

https://github.com/libratbag/piper/blob/d55f0880d9c7d4bae540941acd194dee177f9923/meson.build#L93

Normal users should not have access to `/usr{,/local}/lib/python3.8/site-packages/piper/`, this means python will need to generate the bytecode every time piper is run. This might not be an issue on normal desktop systems, but have you tried to run a python script on a lower-end device like a raspberry pi? This also creates another issue which is if the user running piper has access to `/usr{,/local}/lib/python3.8/site-packages/piper/`, an untracked `__pycache__` directory will be created.

stephanlachnit:
> Actually, it seems like we are not installing __pycache__ for the normal piper:

Ah yeah you're right. Then it's actually not that important (I still prefer this instead of adding it to my gitignore though).

> This might not be an issue on normal desktop systems, but have you tried to run a python script on a lower-end device like a raspberry pi?

I don't want to say that it's useless, I just don't think it's really necessary here. Besides, the speed of the code doesn't change, bytecode is still created, just on every startup.

FFY00:
> Ah yeah you're right. Then it's actually not that important (I still prefer this instead of adding it to my gitignore though).

Then `python -B piper.devel` instead of `./piper`.

> I don't want to say that it's useless in general, I just don't think it's really necessary here. Besides, the speed of the code doesn't change, bytecode is still created, the compilation time is just added to the startup time.

Either way, we should definitely install `__pycache__` on `ninja install`

stephanlachnit:
> Either way, we should definitely install `__pycache__` on `ninja install`

Please don't. First of all it doesn't make a notable difference, and second of all this would make the whole process super complicated. First of all ninja doesn't generate bytecode, so the build process would need to execute Piper. Secondly, python bytecode is arch specific. This maybe isn't a problem on most distros that only support amd64, but there are distros like Debian which try to support all major architecture like power, arm, risc-v etc, and this would make the whole process more complicated for basically no advantage.

FFY00:
> Please don't. First of all it doesn't make a notable difference, and second of all this would make the whole process super complicated. First of all ninja doesn't generate bytecode, so the build process would need to execute Piper. Secondly, python bytecode is arch specific. This maybe isn't a problem on most distros that only support amd64, but there are distros like Debian which try to support all major architecture like power, arm, risc-v etc, and this would make the whole process more complicated for basically no advantage.

Python bytecode is definitely not arch specific. You can think of bytecode as pre-parsed python files. `.pyc` and `.pyo` can be easily removed by packagers but we can include a build system switch for that. Like I said, if we don't install those files, they will be created automatically at runtime, making them untracked. This makes it so that meson is unable to unistall them, and the same goes for package managers if there isn't a specific mechanism in place to deal with this. Debian/Ubuntu have post and pre install scripts to remove and add the files, respectively.

As an upstream we maintain the `install` and `uninstall` routines, in this case, our `unistall` behavior is incorrect.

whot:
I think we can close this one now, agreement is that we're not setting that value.

PR #505, opened 2020-04-19
Implement disabling a button

### LINKED ISSUE #247: Provide an option to unbind a button
See https://github.com/libratbag/piper/issues/241#issuecomment-416886673 and the following comment.

### LINKED ISSUE #498: Unassign buttons / Set action for button to nothing
So this is the only ttol which lets me change the behaviour of my forward / backward buttons. However, it seems to me at least, you HAVE to set a definitve action for a button. Right now I have set to to normal mouse click. Would it be possible that the pressed button does nothing , meaning unassign complete action from the button.

stephanlachnit:
Based on #500, closes #247.

Not tested with hardware yet. Probably doesn't work without throwing an error (see commented line). Looks ugly (adds new section), but I don't see an easy method without a bigger code refactoring, for which I don't have the time rn.

### DIFF
--- a/piper/buttondialog.py
+++ b/piper/buttondialog.py
@@ -97,6 +97,8 @@ class ButtonDialog(Gtk.Dialog):
         self._current_macro = None
         self._button = ratbagd_button
         self._action_type = self._button.action_type
+        if self._action_type == RatbagdButton.ActionType.NONE:
+            self._mapping = self._button.disabled
         if self._action_type == RatbagdButton.ActionType.BUTTON:
             self._mapping = self._button.mapping
         elif self._action_type == RatbagdButton.ActionType.MACRO:
@@ -152,6 +154,11 @@ class ButtonDialog(Gtk.Dialog):
             if self._action_type == RatbagdButton.ActionType.SPECIAL and key == self._mapping:
                 self.listbox.select_row(row)
             i += 1
+
+        row = ButtonRow(_("Disable"), _("Misc mapping"), RatbagdButton.ActionType.NONE, 'disabled')
+        self.listbox.insert(row, i)
+        if self._action_type == RatbagdButton.ActionType.NONE:
+            self.listbox.select_row(row)
 
         if self._action_type == RatbagdButton.ActionType.MACRO:
             self._create_current_macro(macro=self._mapping)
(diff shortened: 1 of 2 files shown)


piper/buttondialog.py
```diff
@@ -153,6 +155,11 @@ def _init_other_buttons_ui(self, buttons):
                 self.listbox.select_row(row)
             i += 1
 
+        row = ButtonRow(_("Disable"), _("Misc mapping"), RatbagdButton.ActionType.NONE, 'disabled')
```
FFY00, line 158: I am not really happy with this, you are creating a new row for `RatbagdButton.ActionType.NONE` which only has and will ever have one action, disable.

stephanlachnit: Me neither, I really don't like it. I tried to add an entry in the glade file similar to the keystroke entry, but that only results in errors. I'm not even entirely sure why, I suspect it this because of [this function](https://github.com/libratbag/piper/blob/master/piper/buttondialog.py#L132).
This function will failing trying to do [this](https://github.com/libratbag/piper/blob/master/piper/buttondialog.py#L194) because `_section` doesn't exist. Easiest workaround: add a Section.

FFY00: I don't think the code was designed with glade. It relies a lot on code.

stephanlachnit:
> I already had a look at this and I think it only makes sense to add disable action to the list if we get rid of the sections.

Agreed, I don't think they add any value anyway, at least for the end users it looks a bit odd anyway (e.g. Button 5 is not special but forward is).
Problem is: you basically need to rewrite the whole thing, and university starts for me this week so it ain't gonna happen soon (before I'd rather invest my time on the DeviceType property in libratbag).

> I think it would be better for us to just add a sub-button to enable/disable the action.

Can you elaborate? Do you mean adding a button somewhere in the dialog (not in the listbox)?

piper/buttondialog.py
stephanlachnit: It uses a mix of both, which I suspect is the cause of my problem. [This](https://github.com/libratbag/piper/blob/master/data/ui/ButtonDialog.ui#L88) is the row for the keystroke.

khimaros:
@stephanlachnit is anything blocking this PR from being submitted? something others could help with?

staticssleever668:
> I think it would be better for us to just add a sub-button to enable/disable the action.

I think current design is better. This would require adding new capability for disabling buttons, because with the way buttons are mapped to NONE (not with https://github.com/libratbag/libratbag/pull/953 merged) it won't be possible to turn a button back on.

PR #525, opened 2020-05-25
meson: remove runtime dependency checks

stephanlachnit:
I want to remove all runtime dependency checks for Piper in Meson, that aren't actually required to build/install Piper.
The reasoning is pretty obvious: the build environment might not be the use environment in a lot of scenarios. For example, in Debian the package is created from a build server, that never runs Piper, however it still needs the Python `gi` module etc, which consumes unnecessary build time and should be avoided.
It also reduces the CI runtimes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,34 +11,52 @@ version_date='2020-06-22'
 ratbagd_api_version = 1
 
 
-# Dependencies
-dependency('pygobject-3.0', required: true)
-ratbagd = find_program('ratbagd', required: false)
-ratbagd_required_version='0.14'
+# Options
+enable_runtime_dependency_checks = get_option('runtime-dependency-checks')
+enable_tests = get_option('tests')
 
-if not ratbagd.found()
-    message('''
-    *********************** WARNING **********************
-    * _???_     Dependency libratbag not found in $PATH! *
-    *(_)_(_)    ratbagd must be installed and running    *
-    * (o o)     for Piper to work.                       *
-    *==\o/==                                             *
-    ******************************************************
-    ''')
-else
-    r = run_command(ratbagd, '--version')
-    if r.returncode() != 0 or r.stdout().version_compare('<' + ratbagd_required_version)
-    message('''
-    *********************** WARNING ****************************
-    *                 (   )(                                   *
-    *         (\-.     ) (  )    ratbagd found in $PATH is too *
-    *         / _`> .---------.  old and  will  not  work with *
-    * _)     / _)=  |'-------'|  this version of Piper. Please *
-    *(      / _/    |O   O   o|  update libratbag/ratbagd.     *
-    * `-.__(___)_   | o O . o |                                *
-    ************************************************************
-    ''')
+# Dependencies
+if enable_runtime_dependency_checks
+    ratbagd = find_program('ratbagd', required: true)
+    ratbagd_required_version='0.14'
+
+    if not ratbagd.found()
+        message('''
+        ************************ ERROR ***********************
+        * _???_     Dependency libratbag not found in $PATH! *
+        *(_)_(_)    ratbagd must be installed and running    *
+        * (o o)     for Piper to work.                       *
+        *==\o/==                                             *
+        ******************************************************
+        ''')
+    else
+        r = run_command(ratbagd, '--version')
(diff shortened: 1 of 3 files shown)


whot:
The reason why we have these checks is because we spent an uncomfortable amount of time closing issues filed by users with the dependencies missing (after installing from git or wherever). 

Having this complain at build time looks like a setup issue, having it fail at runtime (which may not even show up if you just click the icon in the GUI) looks like a bug. So the bug gets filed and you get to triage it and close it after telling the user to install the dependencies. You waste more time and resources than the debian builders ever will...

stephanlachnit:
@whot new idea: what about adding a build option to remove the dependency checks which is off by default? This way we still show the error during build to the user, but also allow building it without any of the additional runtime dependencies.

FFY00:
To me, this just feels like we are adding a bunch of options willy nilly. But if it does make a significant difference in your use case, I guess I'm fine.

whot:
@stephanlachnit you're optimising to make the computers work less and the humans work more. This is the wrong way round :)

let the computers suffer the build deps, it really doesn't matter because if you get even one bug you have to triage because of wrong dependencies you've spent more time than you ever saved with this option.

stephanlachnit:
> To me, this just feels like we are adding a bunch of options willy nilly. But if it does make a significant difference in your use case, I guess I'm fine.

> @stephanlachnit you're optimising to make the computers work less and the humans work more. This is the wrong way round :)
> 
> let the computers suffer the build deps, it really doesn't matter because if you get even one bug you have to triage because of wrong dependencies you've spent more time than you ever saved with this option.

Well maybe I should elaborate a bit more why I want this:
When packaging stuff for Debian, one usually does it in a clean chroot to avoid any changes from potential user configs. This means, nothing but the most essential packages are installed, and any package required for building are installed on the fly. Even with already downloaded packages being cached, this takes a couple of minutes for PyGObject.
On projects that actually compile something, I don't really mind, but in this case most of the build time is simply not required. When I want to change something in package (i. e. some testing), this becomes really annoying.

The way I designed it right now is basically equivalent to before. All the runtime dependency checks are run by default. However, one can now disable these checks. A user that doesn't look at the build instructions in the wiki will also never look into the build options, so your original point doesn't apply anymore (which I btw totally agree with).

And just in case someone does actually build from source, disable the checks and doesn't have the dependencies installed, I will triage the bug with please ;)

meson_options.txt
```diff
@@ -2,3 +2,8 @@ option('tests',
 	type: 'boolean',
 	value: true,
 	description: 'Run the tests (default=yes)')
+
+option('runtime_depcheck',
```
whot, line 6: `runtime-dependency-checks` - less ambiguous and using underscores in options is generally frowned upon

meson.build
```diff
@@ -11,34 +11,52 @@ version_date='2020-06-22'
...
+        *********************** WARNING ****************************
+        *                 (   )(                                   *
+        *         (\-.     ) (  )    ratbagd found in $PATH is too *
+        *         / _`> .---------.  old and  will  not  work with *
+        * _)     / _)=  |'-------'|  this version of Piper. Please *
+        *(      / _/    |O   O   o|  update libratbag/ratbagd.     *
+        * `-.__(___)_   | o O . o |                                *
+        ************************************************************
+        ''')
+        endif
```
whot, line 45: in two minds about this. I'm wondering whether to make this required, after all we have the  dependency-check option now.

stephanlachnit: Sounds reasonable, but how should we handle it if the version is smaller than `ratbagd_required_version`? Aborting or just a warning? When thinking about bug triaging, aborting seems like the better idea.

whot: abort. If the user builds piper but has an old ratbagd version running, you're guaranteed to blow up later anyway, might as well make this look more intended.

The chance of the user restarting into a different ratbagd before starting this compiled piper version is so minuscule as to be statistical noise.

stephanlachnit:
Oh wow I just noticed that the CI completely ignores the `MESON_PARAMS` variable, which means we always install into `/usr/local` and have actually never build with `buildtype=release` or `buildtype=plain`.

stephanlachnit:
Worked around the variable problem by adding `-Druntime-dependency-checks=false` just all the time. Still, the problem persists and I tracked it in #603.

PR #527, opened 2020-05-25
Use SPDX 3 license identifiers in the code

stephanlachnit:
SPDX license identifiers make it easier to find out under what license a file is.

Obviously, this will only work if @Hjdskes (current copyright owner) agrees to that change.
Also, I want to put `piper/svg.py` under the GPL-2.0 like the rest of the code. This file was entirely written by @whot, so that only works if you agree as well.

### DIFF
--- a/piper/application.py
+++ b/piper/application.py
@@ -1,18 +1,4 @@
-# Copyright (C) 2017 Jente Hidskes <hjdskes@gmail.com>
-#
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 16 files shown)


whot:
am I missing something here? the `svg.py` file is already under GPLv2+, it's the same license blurb as in the few other files I opened to check.

Copyright is funny, in that what's in the actual file doesn't really matter. Most ppl never update the `Copyright $year $author` lines after the first few commits to a file and what is copyrightable is a tad blurry anyway. e.g. adding device IDs to a file is probably not copyrightable . 

But anyway, who has copyright on a file is in the git log, not i the `Copyright ` header lines.
In my non-lawyery opinion, you are not relicensing things, you're merely changing the way the license is referenced. I don't think you technically need my or @Hjdskes permission for that change. And an option to avoid that completely is to simply add the line to the top without changing the existing comments.

All that to say: you have my permission to add the SPDX license headers in lieu of the fully license comments.

Hjdskes:
I concur with @whot, i.e., in so far that it is relevant, you have my permission to add the SPDX license headers.

PR #528, opened 2020-05-25
Use Gtk's Template class

### LINKED ISSUE #219: PyGI templates support is officially supported now
Someone was asking about Gtk.Builder templates support in Python. When referring this person to my summary of the unofficial `gi_composites.py`, I found that recently official template support was added to PyGObject. See https://gitlab.gnome.org/GNOME/pygobject/merge_requests/52.

If this supports all our needs, we might be able to drop the unofficial `gi_composites.py` for newer versions of Piper. Less unofficial hacks == more goodness.

stephanlachnit:
Closes #219.


stephanlachnit:
Draft, since it doesn't work (yet):
```sh
$ ./builddir/piper.devel
  File "/home/stephan/Projects/piper/piper/piper/application.py", line 55, in do_activate
    window = Window(self.init_ratbagd, application=self)
  File "/home/stephan/Projects/piper/piper/piper/window.py", line 47, in __init__
    self.init_template()
TypeError: <lambda>() missing 1 required positional argument: 's'
```

stephanlachnit:
`init_template()` doesn't need to be called, but it should work with https://gitlab.gnome.org/GNOME/pygobject/-/commit/94f32d1f0861b66bead712313e8b39f58529f139 (since that isn't in a release yet it doesn't work).

While the code executes now, it doesn't work because for example, the LED Dialog looks entirely different and isn't working. I don't know why though.

staticssleever668:
Superseded by #836.

PR #548, opened 2020-07-04
meson: make tests optional

stephanlachnit:
It should be possible to skip tests. Taken from
https://github.com/libratbag/libratbag/blob/e55b04ec7d41e4566ad585174ef6db108cb2804d/meson.build#L364-L365

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -89,28 +89,30 @@ svgs = [
 
 svg_mapping = files('svgs/svg-lookup.ini')
 
-test_svg_files = find_program('svgs/check-svg.py')
-test('check-svg', test_svg_files,
-     args: svgs,
-     env: ['BASEDIR=@0@'.format(meson.current_source_dir())])
+if enable_tests
+     test_svg_files = find_program('svgs/check-svg.py')
+     test('check-svg', test_svg_files,
+          args: svgs,
+          env: ['BASEDIR=@0@'.format(meson.current_source_dir())])
 
-test('svg-lookup-check',
-     find_program('svg-lookup-ini-test.py'),
-     args: [svg_mapping,
-            join_paths(meson.source_root(), 'data/svgs/')
-     ],
+     test('svg-lookup-check',
+          find_program('svg-lookup-ini-test.py'),
+          args: [svg_mapping,
+               join_paths(meson.source_root(), 'data/svgs/')
+          ],
 
-)
-# meson requires that we specify all files one-by-one, so let's add a test
-# that we can't forget about that.
-test('svg-files-in-meson.build',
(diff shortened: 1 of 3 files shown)


whot:
I don't quite understand why this is needed? In libratbag the tests have extra dependencies and fails the build if that one is missing. Here only `flake8` is needed for one test and it was already optional. The tests are only run during `ninja test` so - why do we need another layer of disabling those?

FFY00:
Oh, yeah. I totally forgot they are run via `ninja test`, :man_facepalm:. Sorry!

stephanlachnit:
> I don't quite understand why this is needed? In libratbag the tests have extra dependencies and fails the build if that one is missing. Here only `flake8` is needed for one test and it was already optional. The tests are only run during `ninja test` so - why do we need another layer of disabling those?

First part of two things. Currently, arch doesn't run flake8: https://app.circleci.com/pipelines/github/libratbag/piper/49/workflows/0c36b3aa-921d-4b31-b44c-941c0e5b25b1/jobs/1047
Obviously it should do that, and I think it should fail if flake8 isn't run, cause else what's the purpose of the tests.
However, I didn't want to remove the `required: false` part _yet_, since the latest flake8 currently fails for `ratbagd.py` (https://github.com/libratbag/libratbag/issues/1017) and I think it makes sense to fix that first (I don't have much time right now else I would have already looked into it).

whot:
> Currently, arch doesn't run flake8 ... what's the purpose of the tests if only half of them are run

It's not quite as black and white. You want CI to run on as many things as possible of course but some tests don't need to. any static source checker will give us the same results regardless of host distribution (not accounting for the occasional hiccups when the version changes and new warnings are added). In most other repos that I maintain flake8 is a separate CI job but here we hooked it up for convenience because a) piper is all python, b) it's not exactly under full-attention maintenance and c) as a general rule no-one but the maintainers look at the CI test results. so I'm not sure I can recommend that, there is something to be said to fail `ninja test` based on flake8 results :)

> I didn't want to remove the required: false part for flake8 yet, since the latest flake8 currently fails for ratbagd.py 

wrong way round: if the tests fail, you fix the code, not disable the test suite :wink:

stephanlachnit:
> You want CI to run on as many things as possible of course but some tests don't need to. any static source checker will give us the same results regardless of host distribution

I guess if the tests were only for the CI, that makes sense, but for example the `check-files-in-git.sh` test will always pass on the CI and is more for the average contributor as far as I see it. In the same way, we want them to run flake8.

> wrong way round: if the tests fail, you fix the code, not disable the test suite wink

True, luckily I never disabled the test suite ;)

Anyway, I fixed it and will open a PR shortly.

PR #550, opened 2020-07-06
meson: add test for appdata file

stephanlachnit:
Currently the test gives an informational warning (doesn't fail):
```
I: org.freedesktop.Piper:3: cid-contains-uppercase-letter org.freedesktop.Piper
    The component ID should only contain lowercase letters.
```

We might want to change the rDNS to `io.github.libratbag.piper`, as it is suggested by https://www.freedesktop.org/software/appstream/metainfocreator. Having a "wrong" url is a bad idea anyway since it might be possible some project on freedesktop.org will call itself Piper and use it as well (unlikely though).
I do heard that changing the component ID is potentially breaking so I'm not sure if we actually want to that (I have no idea though, but I can imagine something like ratings in the store).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,9 +2,9 @@
 
 piper_references:
   build_dependencies: &build_dependencies
-    FEDORA_RPMS: meson gettext python3 pygobject3-devel python3-lxml libratbag-ratbagd python3-cairo python3-evdev python3-flake8 gtk-update-icon-cache
-    UBUNTU_DEBS: meson pkg-config gettext python3 python-gi-dev python3-lxml python3-evdev gir1.2-rsvg-2.0 python3-gi-cairo flake8 ratbagd gtk-update-icon-cache
-    ARCH_PKGS: meson libratbag python-cairo python-evdev python-gobject python-lxml gtk-update-icon-cache flake8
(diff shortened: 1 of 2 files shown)


stephanlachnit:
Maybe also worth doing is checking the desktop file via `desktop-file-validate`

stephanlachnit:
@whot @FFY00 anyone some input about the component ID change?

whot:
> @whot @FFY00 anyone some input about the component ID change?

Simply said - I don't know, sorry. Would have to read up on this and your googling is likely as good as mine ;)

stephanlachnit:
Anyway, we can still add this test without changing the component ID.

stephanlachnit:
> @whot @FFY00 anyone some input about the component ID change?

Quick update: they actually thought of that in the spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides

>     <id/>
> Contains the component-ID of another software component. The presence of this tag indicates that the software component containing it is able to provide all functionality of the one referenced in the [provides-id] tag.
> This is useful in case a component-id had to be renamed in the past, e.g. because its domain-name changed.

whot:
@stephanlachnit - is this one good to go? I just found this PR under the carpet where no-one noticed it for over a year ;)

stephanlachnit:
> @stephanlachnit - is this one good to go? I just found this PR under the carpet where no-one noticed it for over a year ;)

Jup, should still be just fine.

whot:
choo choo, here comes the merge train :steam_locomotive:

PR #555, opened 2020-08-06
Add support for the Logitech G513 Carbon Keyboard

owendaprile:
I created an SVG for my Logitech G513 Carbon and added it's USB identifier. libratbag doesn't support all of the keys that are shown in LG G Hub for it as seen here:
Capture
But the SVG has support for all 12 function keys so they should be automatically picked up when libratbag supports them. It works perfectly for the LEDs.
I don't own the other variants of the keyboard so I don't know if their USB identifiers are different, but they all look the same so they should be fine when added.

Here's some screenshots:
image
image
image
image
image

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -60,6 +60,7 @@ svgs = [
     'svgs/logitech-g500.svg',
     'svgs/logitech-g500s.svg',
     'svgs/logitech-g502.svg',
+    'svgs/logitech-g513.svg',
     'svgs/logitech-g600.svg',
     'svgs/logitech-g602.svg',
     'svgs/logitech-g603.svg',
--- a/data/svgs/logitech-g513.svg
+++ b/data/svgs/logitech-g513.svg
@@ -113,7 +113,7 @@
          ry="0.48236895" />
       <rect
          style="fill:#eeeeec;fill-opacity:0.941176;stroke:#babdb6;stroke-width:0.252251;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-         id="button0"
+         id="button0_DISABLED"
          width="4.5405092"
          height="4.5405111"
          x="16.698566"
@@ -122,7 +122,7 @@
          inkscape:label="F1" />
       <rect
          style="fill:#eeeeec;fill-opacity:0.941176;stroke:#babdb6;stroke-width:0.252251;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
-         id="button1"
+         id="button1_DISABLED"
          width="4.5405087"
          height="4.5405111"
          x="21.743578"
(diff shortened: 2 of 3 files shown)


FFY00:
Thank you! We are admittedly a bit behind on adding SVGs for the devices.

Can you confirm you can actually remap the function keys?

owendaprile:
No, it wants to set up buttons 0 and 1 like left and right mouse buttons
image
and the options for 2 and 3 are like the options for other mouse buttons.
image

Trying to set it to a macro results in this error
image
and applying it doesn't save any changes.

Setting one of the preset options saves but it doesn't actually do anything when pressing on the keyboard.
image

I realized I can just comment out the whole Buttons layer in the SVG until support is added for remapping the keys which keeps it working for LEDs. Let me know if I should commit that.
image
If there's any sort of guide for reverse engineering the communication I can take a look and try to add support to libratbag.

FFY00:
Yeah, we need to fix that. Keyboard support is recent, we used to only have mice.

There is some documentation about the protocol [here](https://github.com/Logitech/cpg-docs/tree/master/hidpp20). It only has the packet structure and the feature (the protocol declares "features", which are basically groups of functions) discovery routine (`IRoot`) right now. For more info about the features and functions available, you can have a look at https://github.com/libratbag/libratbag/blob/master/src/hidpp20.c where a lot of them are implemented.

Documentation about the `0x1b04` feature, which the G513 uses for button remapping is available [here](https://lekensteyn.nl/files/logitech/x1b04_specialkeysmsebuttons.html). It needs to be rewritten before going into the the https://github.com/Logitech/cpg-docs repo. Some alternative documentation about the protocol structure and some features is also available [here](https://lekensteyn.nl/files/logitech/logitech_hidpp_2.0_specification_draft_2012-06-04.pdf).

To capture the communication you can use wireshark. If you want to easily filter the data, you can build wireshark with https://code.wireshark.org/review/#/c/37831/ (also available at https://github.com/FFY00/wireshark/tree/hid) and use the `usbhid.data.vendor`. A tip: in wireshark, right click on the data field and select "Apply as a Column".

owendaprile:
I can try and take a look at that soon. For now, should I just comment the Buttons layer so there's at least an image for LED customization?

FFY00:
Yes, please remove the button IDs/labels while we don't know which buttons are actually configured.

owendaprile:
I don't think I'm going to get it to pass check-svg.py since it always wants a minimum of 3 buttons but I don't see any issues in the UI here.

FFY00:
This will have to be stalled then :confused:

doraskayo:
@FFY00, is there any way to get this un-stalled?

FFY00:
Hey, I forgot to ping this PR but #597 should have disabled this check. Could you please rebase on master? Thanks!

owendaprile:
Wow sorry for the delay, GitHub just never notified me about the reply. Anyway, this should be good now. If me or anyone else gets button remapping working for this, then these need to be changed for the buttons to show again:

* The IDs for buttons 0-11 need to have `_DISABLED` removed from them,
* and the block from line 1049 to 1333 must be uncommented.

owendaprile:
Bumping this as it should still be good to merge

owendaprile:
@FFY00 Any update on this? There haven't been any new commits so it should be fine to merge.

FFY00:
Sorry for the delay, I have removed the merge commit from you branch and rebased it on the current master instead, to avoid pinging you again.

FFY00:
Merged, thanks :blush:

Issue #322, opened 2019-03-17
cant set button to esc key


5af64758:
the dialog for setting a macro keystroke closes when you press esc, meaning you cant set any of your buttons to esc

Bensge:
I would like to bring attention to this issue. I usually map one mouse button to the escape key, but I am unable to do so with Piper

whot:
This is a limitation of the key macro recording we have now, shouldn't be a limitation of the underlying software. If you can, please help us implement this.

scresante:
Same problem applies to macro involving the enter key

jeffkarney:
Is there any reason removing this chunk of code from _do_key_event in piper/buttondialog.py would cause issues? There is already an Apply and Cancel button that accomplishes the same think Enter and Escape currently do.

```py
            # Return accepts the current keystroke.
            if event.keyval == Gdk.KEY_Return:
                self._current_macro.accept()
                return
            # Escape cancels the editing.
            elif event.keyval == Gdk.KEY_Escape:
                if self._action_type == RatbagdButton.ActionType.MACRO:
                    self._create_current_macro(macro=self._mapping)
                else:
                    self._create_current_macro()
                self.stack.set_visible_child_name("overview")
                return
```
I have done this locally and was able to use both ESC and Enter in my button macros. I haven't seen any issues caused by it.

whot:
The only "issue" would be that you can't cancel editing with Esc anymore :) I think it's an acceptable compromise given the desire to map Esc, so green light from me.

fctorial:
When will this feature be available. I'm using piper 0.5.1 and can't bind buttons to escape key.

stephanlachnit:
Until someone will do a new release. We track interest here: https://github.com/libratbag/libratbag/issues/974

Without more reviews, nothing will happen soon, sorry.

fctorial:
For others who want this feature, [this](https://github.com/fctorial/dotfiles/blob/master/nix-config/piper_enter_esc.patch) patch can be applied to piper 0.5.1. For nixos users, just install 

    piper.overrideAttrs (oa: {patches = [./piper_enter_esc.patch];})

instead of `piper` and you're good.

Azarilh:
> For others who want this feature, [this](https://gist.github.com/fctorial/60fcda2489a17fb7f78ca1b37b9a0090) patch can be applied to piper 0.5.1. For nixos users, just install

How to apply the patch?

PR #565, opened 2020-08-30
Don't close/accept macro dialog on esc/enter to allow binding these keys

AlexP11223:
#322


Cxarli:
Could I please bump this? My current workaround is very dirty and this would be a lifesaver.

FFY00:
@stephanlachnit do you have some time to look at this?

stephanlachnit:
I can take a look at it today.

stephanlachnit:
Yep works, code is fine as well

PR #578, opened 2020-09-26
data: add SVG for the Logitech G915

opal06:
The SVG is reused from the Logitech G815, as it is the same device, just wireless.


FFY00 (changes requested):
Sorry for the delay, you can just use the G815 SVG directly, no need to make a copy.

opal06:
Ok, have changed it to use the g815 svg. Should be fine now. Please let me know if you any additional changes are needed :)

FFY00 (approved):
Yeah, it is fine now :blush:

PR #588, opened 2020-10-19
translation: fix and improve German strings

NicoWeio:
I found a number of errors in the German translation and fixed them. There are a few opinionated changes, too.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -9,7 +9,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-01-28 20:40+0100\n"
 "PO-Revision-Date: 2017-08-23 13:02+1000\n"
-"Last-Translator: Peter Hutterer <peter.hutterer@who-t.net>\n"
+"Last-Translator: Nicolai Weitkemper <kontakt@nicolaiweitkemper.de>\n"
 "Language-Team: German <>\n"
 "Language: de\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


po/de.po
```diff
@@ -323,14 +323,15 @@ msgstr "Besuche Piper auf GitHub"
 #: data/ui/AboutDialog.ui.in:13
 msgid "translator-credits"
 msgstr "Peter Hutterer <peter.hutterer@who-t.net>, 2017-2019"
+"\nNicolai Weitkemper <kontakt@nicolaiweitkemper.de>, 2020"
```
NicoWeio, line 326: I'm not sure if it's appropriate to add myself to the list of contributors for these small corrections. I also have no idea whether a newline is right here.

po/de.po
```diff
@@ -9,7 +9,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-01-28 20:40+0100\n"
 "PO-Revision-Date: 2017-08-23 13:02+1000\n"
-"Last-Translator: Peter Hutterer <peter.hutterer@who-t.net>\n"
+"Last-Translator: Nicolai Weitkemper <kontakt@nicolaiweitkemper.de>\n"
```
NicoWeio, line 12: Not sure if/how to update these.

po/de.po
whot: mostly depends on whether you want the claim to fame :) this is imo a leftover from early approaches where all this was only tracked in the source files and not in other trackers like github.

The newline should be there iirc, just check the About dialog for correctness. But isn't this bit autogenerated anyway?

po/de.po
whot: depends on whether you want the credit. these days we look at github more so than these files, but having a contact to someone who can do the translations is always good.

whot (approved):
looks good to me and sorry about the embarrassing errors here. in high-school that would've been a "Fünfer" :) Anyway, lemme know if you want to remove your email after all, I don't mind either way.

po/de.po
NicoWeio: The newline looks fine. These lines weren't auto-generated for me.

po/de.po
NicoWeio: Okay, I'll leave it in then. :)

NicoWeio:
Well, if you don't consider it bad form, I'd like to leave them in. I'm always happy to extend my references. :)

whot:
merged, thanks

Issue #590, opened 2020-11-01
New app icon?


bertob:
The current app icon doesn't really follow the HIG, would you be intersted in help with a new one?

Here's a quick experiment redrawing the current metaphor in the platform style:

![image](https://user-images.githubusercontent.com/1908896/97807334-77f64b00-1c60-11eb-8e17-80de337772a6.png)

FFY00:
Yeah, definitely. I think that icon looks good.

@whot @stephanlachnit thoughts?

stephanlachnit:
Love it

whot:
heh, yes, very nice. and definitely better than my 10 minute icon ;)

bertob:
Alright, I'll clean it up a bit and make a PR then :)

PR #594, opened 2020-11-11
app icon: update style, add source file, add symbolic

bertob:
As discussed in #590

image

This just adds the new assets, and doesn't hook up the symbolic to the build system or switch the code to using the actual app icon instead of the 404 one.


stephanlachnit:
Thanks again. Any reason the 404 got deleted?

Anyway, the build files need to be updated, but if you want I can also do it.

bertob:
> Any reason the 404 got deleted?

I deleted it because it's the old icon, so it'll have to be replaced with the new one somehow. I was also thinking it might be better to avoid duplicate assets by just using `org.freedesktop.Piper.svg` here.

stephanlachnit:
Sorry for the long waiting period. I wired up Meson support in #602. Currently we can't remove `404.svg`, would require some bigger changes I don't have time for.

Thanks again for the new icon! If there are no last objections, I would merge #602.

bertob:
Sounds great!

PR #597, opened 2020-11-25
tests: check-svg: have no minimum required button count

FFY00:
I was hoping to be able to come around to this and refactor the test to introduce device types, but I haven't been able to. So just disable the minimum button count so that we can move #555 forward.


whot:
merged, thanks

PR #577, opened 2020-09-24
Add sample svg for mx vertical

faridgh1991:
sample svg for MX Vertical added, it is copy of g403's svg that i removed extra buttons.we need to replace it by a correct svg. #539

PR #606, opened 2020-12-16
data: add mx vertical support

sketchbuch:
Adds support for Logitech MX Vertical

I have built piper and it runs. I can change the buttons but, if I apply my changes, piper shows a notification saying that something went wrong and that buttons have been restored to their previous state. Is anyone able to tell me why? Is it because it wasn't the real app, just the "./builddir/piper.devel" version? This also happens with the current version of piper on the 404 svg so doesn't seem to be something my pr has caused

Also are there any docs on what the SVG LED layer is for? Currently I left it empty. Is it the blue highlights you see when selecting a mouse button function as seen in the fallback SVG?

Changes to "data/svgs/svg-lookup.ini" were taken from a similar PR
https://github.com/libratbag/piper/pull/577

My SVG is based on an overhead photo* of a MX vertical and looks more like the real mouse than the SVG in 577

cc: @faridgh1991 

*https://media.officedepot.com/image/upload/b_rgb:FFFFFF,c_pad,dpr_1.0,f_auto,h_1665,q_auto,w_1250/c_pad,h_1665,w_1250/v1/products/4057397/4057397_o01_logitech_mx_vertical_wireless_mouse_1?pgw=1&pgwact=1

image


FFY00:
This looks great, thanks! The LED layer is for devices with addressable LEDs, which the MX Vertical doesn't have. It is usually gaming devices that have configurable LEDs :blush:

The only feedback I have is to please remove the Logitech ("logi") logo. Everything else is fine, but we don't want to have the logo for legal reasons.

sketchbuch:
I have removed the logo.

Still having the problem with Piper that there is an error after applying changes and then all the changes are reverted. Is this something you can fix?

youduda:
> Still having the problem with Piper that there is an error after applying changes and then all the changes are reverted. Is this something you can fix? It isn't just with the test build, this also occured with the latest version of Piper.

I have the same issue if a special mapping is assigned to one of the buttons (not supported see here: libratbag/libratbag#830). So as long as none of the special mappings is used, everything can be configured properly.
The remaining problem for me is that every time the mouse is turned off and on, the configuration is reset.

FFY00 (approved):
Thank you! And sorry for the delay.

PR #632, opened 2021-04-04
Add svg for Logitech Performance Mouse MX

strangemax:
I made a svg for this older mouse.

Unfortunately I'm really unsure about the button mapping. Obviously they are different to the buttons a 'xev | grep button' tells me...

### DIFF
--- a/data/svgs/logitech-performance-mouse-mx.svg
+++ b/data/svgs/logitech-performance-mouse-mx.svg
@@ -1,0 +1,468 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


whot:
The test case failures should be fixed with #711

There's one line (third one from bottom) that doesn't seem to be pointing at a button? Or is that just hard to see?

Otherwise the buttons are numbered in the physical device order, xev will give you button 8 for physical button 4, button 9 for phys 5, etc.

strangemax:
Thanks for the reply. 

The lines to the buttons should be all fine. The dubious button could be seen here:
https://user-images.githubusercontent.com/11463298/148992224-94981865-a506-4231-a656-1521da78fe21.png

Unfortunately I've no time to test it at the moment. Give me a couple of weeks...

PR #685, opened 2021-10-09
Fix icon rendering in the Qt SVG renderer

noahadvs:
The Qt SVG Renderer only supports the SVG Tiny spec, so some features of SVG cannot be used without causing graphical glitches such as black backgrounds or certain effects simply not being rendered.

I've double checked that the appearance of the icons have not changed. The SVG code I removed wasn't for the visible parts of the icons. A side effect of this change is that the SVGs are a bit more optimized now.

Screenshot_20211008_202533
Screenshot_20211008_202716


noahadvs:
Actually, never mind. I only tested it in the Qt SVG renderer and now it looks broken in Firefox. I don't have time to figure this out right now, so I'm closing this MR. The symbolic icon should be OK, but it's not really seen anywhere.

Issue #674, opened 2021-09-02
Add polling rate 125 & 250 Hz


paul-canada:
Piper show only 500 & 1000 Hz polling rate but **Logitech G300** has 125, 250, 500 & 1000.
 
I tried with ratbagctl Logitech rate set 250 or 150. 
after It indicates 200 or 100 but when I tested on website it is still 500 Hz.
I don't want use too much CPU with high polling rate like 500 or 1000 hz. 

Without this issue, Piper works well on ubuntu 20.04 LTS. 
CTRL et SHIFT doesn't work alone for Macro buttons.

PR #703, opened 2021-12-23
add 125 and 250 rates

kyokenn:
Fixes https://github.com/libratbag/piper/issues/674

### DIFF
--- a/data/ui/ResolutionsPage.ui
+++ b/data/ui/ResolutionsPage.ui
@@ -61,17 +61,48 @@
                         <property name="homogeneous">True</property>
                         <property name="layout_style">expand</property>
                         <child>
+                          <object class="GtkRadioButton" id="rate_125">
+                            <property name="label">125</property>
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="receives_default">False</property>
+                            <property name="draw_indicator">False</property>
+                            <property name="group">rate_250</property>
+                          </object>
+                          <packing>
+                            <property name="expand">True</property>
+                            <property name="fill">True</property>
+                            <property name="position">0</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkRadioButton" id="rate_250">
+                            <property name="label">250</property>
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="receives_default">False</property>
+                            <property name="draw_indicator">False</property>
+                            <property name="group">rate_125</property>
+                          </object>
+                          <packing>
+                            <property name="expand">True</property>
+                            <property name="fill">True</property>
+                            <property name="position">1</property>
+                          </packing>
+                        </child>
+                        <child>
                           <object class="GtkRadioButton" id="rate_500">
                             <property name="label">500</property>
                             <property name="visible">True</property>
                             <property name="can_focus">True</property>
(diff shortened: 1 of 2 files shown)


paul-canada:
How I can apply this fix ? 
Thanks.

kyokenn:
> How I can apply this fix ? Thanks.

You can wait until it get reviewed by maintainers and merged, or you can patch it by yourself or download my repository.

paul-canada:
Yes, I can wait the upgrade.
1. How I can patch ? 
2. I downloaded also your file piper-master.zip (23 Dec). How upgrade piper 0.6 ? 
Thanks for your help.

kyokenn:
> I downloaded also your file piper-master.zip (23 Dec). How upgrade piper 0.6 ?

Now you have to unpack, build and install it or you can try to replace those 2 files that I have changed where you have already installed piper.

paul-canada:
> Now you have to unpack, build and install it or you can try to replace those 2 files that I have changed where you have already installed piper.
I failed to compile because my ratbagd is too old and for patch I did not find the file ResolutionsPage.ui 

I think I will wait the "official" update. 
Thanks for your Fixes. :)

paul-canada:
> Now you have to unpack, build and install it or you can try to replace those 2 files that I have changed where you have already installed piper.
    
I failed to compile because my ratbagd is too old and for patch I did not find the file ResolutionsPage.ui

I think I will wait the "official" update.
Thanks for your Fixes. :)

Ablinne:
I have just tried this patch and there seems to be a missing UI update when switching to a profile (via profile chooser in the top left) that uses one of the new refresh rates. The UI stays on the refresh value from the previously selected profile.

kyokenn:
>I have just tried this patch and there seems to be a missing UI update when switching to a profile

I can't reproduce that. Make sure that you are testing with "piper.devel" instead of "piper", otherwise you will get imported the default system-wide installed piper modules.

Also did you compared the real rate values as reported by ratbagctl <device> info ?

Ablinne:
I installed it system wide by modifying the PKGBUILD from the AUR package piper-git (changed only source to this branch, left everything else as it was). When restarting piper, the correct rate was shown again. I didn't try ratbagctl though. I'm out of time for today, I hope that I manage to revisit this tomorrow of the day after.

whot:
Merged, thanks! I'd appreciate a follow-up to check if those resolutions are actually available on the device (they are exported as the `ReportRates` property) and only showing them conditionally.

kyokenn:
>I'd appreciate a follow-up to check if those resolutions are actually available on the device (they are exported as the ReportRates property) and only showing them conditionally.

Nice. I didn't know about it. I thought it was a planned feature (exposing list of supported options) - https://github.com/libratbag/libratbag/issues/1170

whot:
that issue relates to device capabilities which is a slightly different circle in the venn diagram :) the report rates are already available as the [`report_rates` property](https://github.com/libratbag/piper/blob/master/piper/ratbagd.py#L490)

Ablinne:
@whot I double checked, G-Hub offers 125 and 250 as rate settings. Even more strange: As long as I don't *set* it using ratbagctl, it is reported correctly as 125
```
alex@galaxy:~$ ratbagctl "Logitech MX518 Gaming Mouse" rate get
125
````
When I set  it once, from then on I only get 100/200.

PR #721, opened 2022-01-21
Replace CircleCI with github actions

whot:
CircleCI is mostly unviable these days because it severely limits the
number of runs we can do. Since we don't do anything fancy anyway, let's
port to github actions, re-using some of the infrastructure libratbag
already has in place.

We use meson/ninja from pip because building with the latest gives us
some heads-up about upcoming changes. And we're now only building on
ubuntu because that's what's easiest in github. Oh well.


whot:
So, perhaps unsurprisingly circleci fails here - webhook is still enabled but the circleci file was removed.

And the actual workflow doesn't run until it's merged (that's github policy for new workflows) but it does run in the fork, so here's the link with glorious results, fsvo of glorious: https://github.com/whot/piper/actions/runs/1726290249

PR #740, opened 2022-04-21
window: enable window icon

staticssleever668:
I don't know if this is a problem with my WM configuration, but I don't have a window icon in the task list, and this patch fixes it.


whot (approved):
well, if it fixes it, let's go with it! :)

PR #753, opened 2022-06-09
Add correct svg for MX Anywhere 3

christiangnrd:
I modified the svg for the Logitech MX Anywhere 3 to look like the Logitech MX Anywhere 3 instead of the MX Anywhere 2. I verified that it works (as much as piper works for this mouse).

One bug as shown in the screenshot below is that the path doesn't show up for button 2 and I wasn't able to fix it.

Hopefully someone can help with that, but even if not, it's still more functional than with the previous SVG (or none).

image

### DIFF
--- a/data/svgs/logitech-mx-anywhere3.svg
+++ b/data/svgs/logitech-mx-anywhere3.svg
@@ -2,20 +2,20 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


whot:
`ERROR:/home/runner/work/piper/piper/data/svgs/logitech-mx-anywhere3.svg:Missing button5-path for button5`

I suspect that's the reason for the path not showing up correctly :)

christiangnrd:
Hi @whot, thanks for the help. The fact that labels and ID in Inkscape are different is what got me. I have fixed the svg and it seems to be working now.

whot:
Merged, thanks!

PR #754, opened 2022-06-10
Better looking buttons for MX Anywhere 2

christiangnrd:
I modified the MX Anywhere 2 left and right buttons to have a better shape.


whot:
Thanks for the patch! I'm having a hard time seeing the difference between old and new though, at least in github's rendering. What was the the actual change here?

whot:
oh, ok, thanks. the github SVG helpers didn't show that highlight, so the two versions looked identical. Merged!

PR #759, opened 2022-06-24
Modern way to handle post-installation actions

xvitaly:
Introduced a new option `-Dscripts` for packaged builds.

Will be very useful in downstream packaging, because Koji/OBS shouldn't run `gtk-update-icon-cache` during the package build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ env:
   UBUNTU_DEPS: |
       pkg-config gettext appstream
       python3 python-gi-dev python3-lxml python3-evdev
-      gir1.2-rsvg-2.0 python3-gi-cairo flake8 gtk-update-icon-cache
+      gir1.2-rsvg-2.0 python3-gi-cairo flake8 gtk-update-icon-cache desktop-file-utils
   PIP_PACKAGES: meson ninja
 
 jobs:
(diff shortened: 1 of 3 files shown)


whot (changes requested):
A quick check of `meson_install.sh` shows it only runs `gtk-update-icon-cache` which is now handled by the [`gnome` meson module](https://mesonbuild.com/Gnome-module.html#gnomepost_install). And that will DTRT, so we don't need an extra options here.

Can you please amend this PR so we use the gnome module instead here? Thanks.

xvitaly:
> Can you please amend this PR so we use the gnome module instead here?

Done.

`gnome.post_install()` is available since meson 0.57 and `update_desktop_database` - 0.59. That's why I added check. On legacy meson versions (Debian stable) the old script will be used.

whot (approved):

whot:
I'd probably argue we can just require meson 0.59 since they're a year old and anyone can get it from pip anyway but meh :) Thanks!

xvitaly:
> I'd probably argue we can just require meson 0.59 since they're a year old and anyone can get it from pip anyway but meh :) Thanks!

I can update the PR if you want.

xvitaly:
Fixed CI.

whot:
> I can update the PR if you want.

Nah, all good, we can drop it in a year or whenever. Thanks for the quick fix, merged!

PR #770, opened 2022-08-05
Replace Quit buttons with normal close buttons

A6GibKm:
Using a Quit button is an anti-pattern and might be confusing for users
expecting the quit button to be on the same position as of other apps'.

Note that the button is on the left and by default the close button is
on the right.

image

### DIFF
--- a/data/ui/ErrorPerspective.ui
+++ b/data/ui/ErrorPerspective.ui
@@ -68,15 +68,7 @@
     <property name="can_focus">False</property>
     <property name="title" translatable="yes">Uh Oh! Something went wrong…</property>
     <property name="has_subtitle">False</property>
-    <child>
-      <object class="GtkButton">
-        <property name="label" translatable="yes">Quit</property>
-        <property name="visible">True</property>
-        <property name="can_focus">True</property>
-        <property name="receives_default">True</property>
-        <signal name="clicked" handler="_on_quit_button_clicked" swapped="no"/>
-      </object>
-    </child>
+    <property name="show_close_button">True</property>
     <style>
       <class name="titlebar"/>
     </style>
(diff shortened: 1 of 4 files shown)


data/ui/ErrorPerspective.ui
```diff
@@ -68,15 +68,7 @@
...
-    <child>
-      <object class="GtkButton">
-        <property name="label" translatable="yes">Quit</property>
-        <property name="visible">True</property>
-        <property name="can_focus">True</property>
-        <property name="receives_default">True</property>
-        <signal name="clicked" handler="_on_quit_button_clicked" swapped="no"/>
-      </object>
-    </child>
+    <property name="show_close_button">True</property>
```
staticssleever668, line 71: Also remove the `_on_quit_button_clicked` method from `piper/errorperspective.py`.

data/ui/WelcomePerspective.ui
```diff
@@ -118,15 +118,7 @@
...
-    <child>
-      <object class="GtkButton">
-        <property name="label" translatable="yes">Quit</property>
-        <property name="visible">True</property>
-        <property name="can_focus">True</property>
-        <property name="receives_default">True</property>
-        <signal name="clicked" handler="_on_quit_button_clicked" swapped="no"/>
-      </object>
-    </child>
+    <property name="show_close_button">True</property>
```
staticssleever668, line 121: Also remove the `_on_quit_button_clicked` method from `piper/welcomeperspective.py`.

staticssleever668 (changes requested):

data/ui/WelcomePerspective.ui
A6GibKm: Done.

whot:
merged, thanks!

PR #795, opened 2022-10-13
Add SVGs for ASUS mice

kyokenn:
Most ASUS mice have the same button layout so I made a two images: one for the mice that have DPI button on top near the scroll wheel and other for mice with DPI button on bottom near the sensor. So those images could be reused later with other ASUS mice, even with less amount of LEDs, because Piper can hide unused SVG layers.


kyokenn:
>Did you mean to close this?

I moved it into separate branch to keep master clean. I can open a new one.

>I think it would be better if you made symlinks from asus-generic-* to mouse names, so for example

It could end up with a lot of symlinks, but if it's ok then I'm fine with it.

PR #796, opened 2022-10-13
resolutionspage: show available rates only

kyokenn:
Show only rate buttons which was reported as supported by mouse.


staticssleever668 (approved):
:+1: 
I thought this would also have to support the case when the mouse reports no available report rates, but it turns out that in this case ratbagd gives  an error about invalid device, so this is fine.

Issue #776, opened 2022-08-13
Introduce Brightness on Solid colors


bozzelliandrea:
Add a brightness slider on the solid color tab, with Logitech G203 solid color can be more lighter.

Basically the cycle or breath functionality without the effect.

PR #797, opened 2022-10-13
leddialog: add brightness to solid led settings

kyokenn:
Fixes https://github.com/libratbag/piper/issues/776


staticssleever668:
While the change itself is good, I don't think we should merge this right now as we would need some changes to libratbag.
One issue is that we would need an analogous change to ratbagctl, which probably won't be hard.
The bigger issue is that we don't support brightness on solid colors on the driver side. As far as I see, this is only supported in `asus`, `sinowealth`, `test` drivers. Because of this having such slider would be very confusing (at least for the time being).

FFY00:
> The bigger issue is that we don't support brightness on solid colors on the driver side.

This is generally implemented as the black component of the color, because you obviously can't have black light.

I had a look, and it seems that `asus` is the only driver which actually sends both brightness and rgb values to the device. It'd be interesting to see what the device does with the black component of the color, and when the color is black.

Anyway, I think that's a special case anyway, we should be able to implement this without any protocol changes, by just using the black component of the color as the brightness. IMO, we should also remove the `brightness` field from the C struct.

PR #804, opened 2022-11-14
implement key bindings feature

kyokenn:
libratbag back-end part - https://github.com/libratbag/libratbag/pull/1368

### DIFF
--- a/piper/buttondialog.py
+++ b/piper/buttondialog.py
@@ -296,13 +296,14 @@ class ButtonDialog(Gtk.Dialog):
 
     def _on_macro_set(self, macro):
         # A macro has been set; update accordingly.
-        if RatbagdButton.ActionType.MACRO in self._button.action_types:  # macro is supported
-            self._action_type = RatbagdButton.ActionType.MACRO
-            self._mapping = macro
-        else:
+        if (RatbagdButton.ActionType.KEY in self._button.action_types and
+                len(macro.keys) == 2):  # single key (press + release events)
             self._action_type = RatbagdButton.ActionType.KEY
             type_, value = macro.keys[0]
             self._mapping = value
+        else:
+            self._action_type = RatbagdButton.ActionType.MACRO
+            self._mapping = macro
         self.stack.set_visible_child_name("overview")
         self._release_grab()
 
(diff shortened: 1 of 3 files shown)


staticssleever668:
Does this make the macro-setting dialog fallback to a single key if full-fledged macros are not supported?
I think instead there should be a new dialog for key actions specifically, with an entry in the action list, just like macro does.

kyokenn:
>Does this make the macro-setting dialog fallback to a single key if full-fledged macros are not supported?

No, I'm just using the first key from the macro.

> I think instead there should be a new dialog for key actions specifically, with an entry in the action list, just like macro does.

I would like to try to reuse the macro dialog and just limit it to single key action.

piper/buttondialog.py
```diff
@@ -294,8 +296,13 @@ def _on_macro_updated(self, macro, pspec):
...
         # A macro has been set; update accordingly.
-        self._action_type = RatbagdButton.ActionType.MACRO
-        self._mapping = macro
+        if RatbagdButton.ActionType.MACRO in self._button.action_types:  # macro is supported
+            self._action_type = RatbagdButton.ActionType.MACRO
+            self._mapping = macro
+        else:
+            self._action_type = RatbagdButton.ActionType.KEY
+            type_, value = macro.keys[0]
+            self._mapping = value
```
staticssleever668, line 305: > I would like to try to reuse the macro dialog and just limit it to single key action.

That does make sense.

Well, then I think this code should use the `key` action type opportunistically, falling back to macro and giving an error message if needed.

Pseudo-code:
```
if SupportsKeyActionType() and macro.keys.len() == 1:
    // SetKeyActionType()
    return
SetMacroActionType()
```

kyokenn:
Changed buttons page so it's now shows key name instead of key code.

staticssleever668:
One problem: when the type of button action is `key`, in the dialog with the list of possible button actions in `Send keystroke` it says `None`.

kyokenn:
> One problem: when the type of button action is `key`, in the dialog with the list of possible button actions in `Send keystroke` it says `None`.

Fixed

staticssleever668 (approved):

PR #836, opened 2023-02-16
Use Gtk's Template class

### LINKED ISSUE #219: PyGI templates support is officially supported now

staticssleever668:
Closes #219.
Replaces non-working #528.

For some yet unknown reason with Gtk.Template the title bar in LED dialog is not shown. I worked around this in d56e7bc5185535ed3d0b8303fa6ea250709c7733, but I would like to know the reason this is needed.


staticssleever668:
I'll merge this with the work-around left in.

PR #838, opened 2023-02-22
Don't show report rate selection box if the device doesn't support it

### LINKED ISSUE #834: Show unchangeable report rate
Logitech Signature M650 mouse shows a changeable report rate:

image

Even though it isn't listed as a feature in the `ratbagctl` output:
```
$ ratbagctl thundering-gerbil info
thundering-gerbil - Logi M650
             Model: bluetooth:046d:b02a:0
 Number of Buttons: 6
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
  Resolutions:
    0: 1000dpi (active)
  Button: 0 is mapped to 'button 1'
  Button: 1 is mapped to 'button 2'
  Button: 2 is mapped to 'button 3'
  Button: 3 is mapped to 'button 4'
  Button: 4 is mapped to 'button 5'
  Button: 5 is mapped to none
```

 Even more bonkers is that if I try to change it, I get an error, and report rate appears in the info output:
```
$ ratbagctl thundering-gerbil info
thundering-gerbil - Logi M650
             Model: bluetooth:046d:b02a:0
 Number of Buttons: 6
    Number of Leds: 0
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
  Report Rate: 500Hz
  Resolutions:
    0: 1000dpi (active)
  Button: 0 is mapped to 'button 1'
  Button: 1 is mapped to 'button 2'
  Button: 2 is mapped to 'button 3'
  Button: 3 is mapped to 'button 4'
  Button: 4 is mapped to 'button 5'
  Button: 5 is mapped to none
```

Using libratbag and piper from current git, with this definition added for the mouse:
```
$ cat data/devices/logitech-Signature-M650.device
[Device]
Name=Logitech Signature M650
DeviceMatch=bluetooth:046d:b02a
Driver=hidpp20
```

`solaar show` doesn't show the mouse as having the 8060 report rate feature.

I think it's possible that Piper assumes that any mouse with changeable DPI has variable report rate, and `ratbagd` doesn't check whether the functionality exists before trying to apply it:
```
ratbag error: hidpp20: failed to update report rate (-95)
ratbagd error: error committing device (-1000)
```

staticssleever668:
Fixes #834.

FIXME: I also noticed that if active report rate is not one of [125, 250, 500, 1000], the button for 125 would still be active. None should be active if possible.

### DIFF
--- a/piper/resolutionspage.py
+++ b/piper/resolutionspage.py
@@ -73,10 +73,10 @@ class ResolutionsPage(Gtk.Box):
             row = ResolutionRow(self._device, resolution)
             self.listbox.insert(row, resolution.index)
 
-        self.rate_125.set_visible(125 in profile.report_rates)
-        self.rate_250.set_visible(250 in profile.report_rates)
-        self.rate_500.set_visible(500 in profile.report_rates)
-        self.rate_1000.set_visible(1000 in profile.report_rates)
+        self.rate_125.set_sensitive(125 in profile.report_rates)
+        self.rate_250.set_sensitive(250 in profile.report_rates)
+        self.rate_500.set_sensitive(500 in profile.report_rates)
+        self.rate_1000.set_sensitive(1000 in profile.report_rates)
 
         # Set the report rate through a manual callback invocation.
         self._on_active_profile_changed(self._device, profile)
--- a/piper/resolutionspage.py
+++ b/piper/resolutionspage.py
@@ -32,6 +32,7 @@ class ResolutionsPage(Gtk.Box):
     rate_250 = GtkTemplate.Child()
     rate_500 = GtkTemplate.Child()
     rate_1000 = GtkTemplate.Child()
+    rate_button_box = GtkTemplate.Child()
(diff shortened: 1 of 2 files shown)


piper/resolutionspage.py
```diff
@@ -12,6 +12,9 @@
 from gi.repository import Gtk  # noqa
 
 
+UNSUPPORTED_REPORT_RATE_SENTINEL = 0
```
staticssleever668, line 15: This is wrong.

Per libratbag API docs: 

ReportRate
uint for the report rate in Hz assigned to this profile. This rate must be one of those listed in [ReportRates](https://libratbag.github.io/dbus.html#ReportRates).

ReportRates
This list may be empty if the device does not support reading and/or writing to resolutions.

staticssleever668: Probably just an oversight in the docs. Not only that, ReportRates may never be empty because the device will fail sanity check otherwise.

Issue #749, opened 2022-05-16
Add debounce time feature


Dexyaa:
Software like the glorious model o has software to change debounce time. It should be implemented here as well

staticssleever668:
Related: libratbag/libratbag#1168.

PR #843, opened 2023-03-03
Add a new page with debounce and angle snapping

kyokenn:
Fixes https://github.com/libratbag/piper/issues/749

### DIFF
--- a/data/piper.gresource.xml.in
+++ b/data/piper.gresource.xml.in
@@ -7,6 +7,7 @@
         <file>svgs/svg-lookup.ini</file>
 
         <file preprocess="xml-stripblanks">AboutDialog.ui</file>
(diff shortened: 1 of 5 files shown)


staticssleever668:
- Rebased to remove merge commit
- Fixed failing flake8 test because of an unused variable
- Reformatted with black (later I'll replace flake8 with it and ruff like I did in libratbag)
- Added type hints where possible (I think it's impossible to type hint children, at least in a clean way ; I'll check if there is a way later)

staticssleever668:
ratbagd.py is now out of sync with libratbag, but it would cause flake8 errors if it was updated now, so it's okay

staticssleever668 (approved):

Issue #470, opened 2020-02-08
Selecting different profile re-activates the 'apply' button


BETLOG:
**Information**
- `ratbagd` version (`ratbagd --version`):
0.12
- Piper version: 
0.3
- Distribution: 
Kubuntu 18.04
- Kernel version (ex. `uname -srmo`): `KERNEL VERSION HERE`
Linux 4.15.0-76-generic x86_64 GNU/Linux

[edit] adding mouse]
- Hardware
Logitech G502 HERO Gaming Mouse

**Describe the bug**
Selecting different profile re-activates the 'apply' button.
This is kind of tedious and can be confusing as to weather the  changes have actually been saved.
If the profile has been changed this is immediately reflected at the mouse, so having an apply function is actually redundant.

[edit] on further issue reading I see that the apply button writes to the mouse's memory....and that doing this often is possibly life-shortening.
Suggest that apply button be removed entirely, and closing piper is the save mechanism. With a "save on exit" checkbox that is checked by default (so we can back out without saving if required).

stephanlachnit:
> Selecting different profile re-activates the 'apply' button.

If I'm not wrong, this is actually intended. If you select a different profile, Piper will mark it as the new active one, and which is not commited to the mouse without hitting the apply button.

> This is kind of tedious and can be confusing as to weather the changes have actually been saved.

Tbh that can be a little bit confusing but I don't see a quick way around this, without the need to entirely redesign the way to choose different profiles. Since Piper sends everything instantly to libratbag via DBus, your changes will always be "saved", and hitting the apply button will always save it on the device, except when you get the error message.

> [edit] on further issue reading I see that the apply button writes to the mouse's memory....and that doing this often is possibly life-shortening.
Suggest that apply button be removed entirely, and closing piper is the save mechanism. With a "save on exit" checkbox that is checked by default (so we can back out without saving if required).

I personally don't think it's that big of a deal, and the apply button is quite nice if you want to test different resolution or LED effects before exiting, so I we are going to remove it.

I'm closing this since I won't change it and the other maintainers don't have a lot of spare time for Piper. If you want to re-design the profile selecting UI go ahead, I would review it.

FFY00:
@stephanlachnit please don't close without getting confirmation from the user if you are not sure about something.

I was testing the G815 yesterday and IIRC the active profile is committed without hitting apply.

stephanlachnit:
[...]

PR #844, opened 2023-03-03
don't trigger apply button while switching profiles

kyokenn:
It's a little bit annoying and weird behavior.
It's explained in https://github.com/libratbag/piper/issues/470
PR fixes this issue.


staticssleever668:
Thank you!
There have been some code style changes on libratbag side, so I will merge it there first to avoid syncing problems. (this file is currently shared between Piper and libratbag code bases)

PR #847, opened 2023-03-03
Refactor modernize code base

staticssleever668:


staticssleever668:
- Adds black and ruff for Python formatting and linting.
- Adds type hints where possible, with the exceptions of tests (didn't have time to look at them yet) and ratbagd.py (I want to find a way to share type hints between it and ratbagc.py in libratbag first).
- Make pages take profiles in the constructors.
This is way easier to maintain and understand, and also may fix some subtle bugs, like `cycle resolution` buttons still being highlighted even after switching profiles, and the `advanced` tab showing/not showing when it's not supposed to (currently on the sinowealth driver).
- Manually disconnect handlers when widgets get destroyed.
This fixes some impossible to diagnose warnings from Gtk and fixes callbacks being called with None when they should've been destroyed.

PR #853, opened 2023-03-05
Fix "commit" button not being active on start even when profile is actually dirty

staticssleever668:


staticssleever668:
This introduces a regression that nothing is ever dirty now :/

PR #856, opened 2023-03-06
Listen and react to D-Bus API proprty changes (for example after ratbagctl)

staticssleever668:
With this Piper will update it's UI to reflect changes to device profile.

### DIFF
--- a/piper/advancedpage.py
+++ b/piper/advancedpage.py
@@ -1,12 +1,15 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+import sys
+from typing import Optional
 import gi
 
 from .mousemap import MouseMap
 from .ratbagd import RatbagdDevice, RatbagdProfile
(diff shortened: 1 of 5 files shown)


staticssleever668:
Not tested profile disabling, but it doesn't work for me at all on sinowealth driver, whose devices don't support profile disabling (or so I think), but the driver doesn't force-enable profiles either. I wonder if this is a regression, I need to test this.

staticssleever668:
The driver was just missing RATBAG_PROFILE_CAP_DISABLE, which makes sense, but ratbagd doesn't give any error what-so-ever in such case. I'll think how this can be improved upon.

staticssleever668:
And yeah profile "deleting" and "adding" (which are actually just disabling/enabling) don't update the UI. However, it doesn't seem like it did before either (checked on d390f22210bd317a58bea1996df6c0b3aba58b3c).

PR #861, opened 2023-03-09
Fix update profile list on disabling

### LINKED ISSUE #859: Profile adding and deleting don't update the profile list
Profile adding and deleting don't update the profile list, and because of that it looks like nothing had happened.
A work-around is to go to the welcome screen and go back. If only a single device is connected to the system, restarting Piper is the only option.

staticssleever668:
Fixes #859.

To be honest, I don't really like the profile delete/add approach, as at the very least it's not the way libratbag works.
Note to myself: check why was it implemented this way instead of disabling/enabling.

### DIFF
--- a/piper/mouseperspective.py
+++ b/piper/mouseperspective.py
@@ -198,11 +198,10 @@ class MousePerspective(Gtk.Overlay):
     def _on_profile_notify_dirty(
         self, profile: RatbagdProfile, pspec: Optional[GObject.ParamSpec]
     ) -> None:
-        if profile is not self._profile:
-            return
+        device_dirty = any(p.dirty for p in self._device.profiles)
 
         style_context = self.button_commit.get_style_context()
-        if profile.dirty:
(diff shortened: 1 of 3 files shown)


staticssleever668:
This needs more work, the button as enabled by default, and also there is some redundant logic in _on_add_profile_button_clicked().

staticssleever668:
There is a problem: pressing the delete button doesn't make the commit button in mouse perspective sensitive, and, by the looks of it, mouse perspective enables the button depending on whether the *currently active* profile is dirty, instead of checking if at least one is dirty.

staticssleever668:
The issue in the last comment is fixed by e5aae89fc7897e32917fb02160e20bf550f69bfc.
A slight drawback of the approach there is that the function body is called as many times as there are profiles on the device. To fix this we would need to emit the `property::dirty` signal on the device instead of the profile, however, I think it's good as-is for now.

PR #868, opened 2023-04-10
Updated Dutch translation

Vistaus:
🙂

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -7,35 +7,32 @@ msgid ""
 msgstr ""
 "Project-Id-Version: piper\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-03-08 20:03+0100\n"
-"PO-Revision-Date: 2020-03-09 13:32+0100\n"
-"Last-Translator: Jente Hidskes <hjdskes@gmail.com>\n"
+"POT-Creation-Date: 2023-04-10 21:12+0200\n"
+"PO-Revision-Date: 2023-04-10 21:26+0200\n"
+"Last-Translator: Heimen Stoffels <vistausss@fastmail.com>\n"
 "Language-Team: Dutch <>\n"
 "Language: nl_NL\n"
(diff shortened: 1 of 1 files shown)


po/nl.po
```diff
@@ -7,504 +7,496 @@ msgid ""
...
+msgid "Other"
+msgstr "Overig"
 
 #. Translators: the {} will be replaced with the button index, e.g.
 #. "Button 1 click".
-#: piper/buttondialog.py:222 piper/buttonspage.py:90
+#: piper/buttondialog.py:257 piper/buttonspage.py:105
 msgid "Button {} click"
-msgstr "Klik met de {}knop"
+msgstr "{}muisklik"
```
staticssleever668, line 90: Does this sound right? It will be like "1muisklik", "2muisklik" and so on.

staticssleever668 (commented):
"DPI" -> "dpi", "LED" -> "led", "API" -> "api", all of these seem like unintended changes

Vistaus:
> "DPI" -> "dpi", "LED" -> "led", "API" -> "api", all of these seem like unintended changes

No, we usually write those terms with non-caps.

po/nl.po
Vistaus: You're right: I may have interpreted that wrong. Will think of a better translation.

staticssleever668 (approved):

staticssleever668:
Thank you!

The "advanced" page seems not to be translated, feel free to do it in another PR.
image

Vistaus:
Hmm, I didn't see the advanced page in nl.po or the pot file. Or at least Poeditor didn't show it to me.

Vistaus:
Thanks for mentioning the advanced and for looking into the issue. :)

staticssleever668:
Fixed: #871.

By the way, when I merged your PR, I force-pushed to your branch, so you will have to do a reset `git reset --hard origin/master`.

Vistaus:
Thanks, I'll do that! :)

PR #876, opened 2023-04-22
Refactory of the code

Domefemia:


staticssleever668:
IMO replacing for loops with next() and replacing early returns with ternary operators make the code less readable (also the walrus operator is Python 3.8+, I think our minimum is 3.6). The rest is good though so I'll just drop these commits.

staticssleever668:
Interestingly, the lines changed in `Converted strings to f string` were not detected by Ruff with UP032, even though I think they should've been. Thanks!

staticssleever668:
Thank you! I'm curious, was this accomplished with a code analyzing tool, or did you just eyeball the code base?

Domefemia:
No problem, sorry for responding late to the various questions, I currectly using pylint to scan the code in combination with black and sourcery to scan the codebase.

PR #887, opened 2023-07-11
Sync the flatpak json

whot:
Our in-tree flatpak file (arguably no longer needed) is well out of date, let's sync it with the flathub one.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -46,7 +46,7 @@ appdata = i18n.merge_file(input: appdata,
                           install: true,
                           install_dir: metainfodir)
 
-appstreamcli = find_program('appstreamcli')
+appstreamcli = find_program('appstreamcli', required: false)
 if appstreamcli.found()
   test(
     'validate appdata file',
--- a/meson.build
+++ b/meson.build
@@ -16,7 +16,7 @@ enable_runtime_dependency_checks = get_option('runtime-dependency-checks')
 
 # Dependencies
 if enable_runtime_dependency_checks
-    ratbagd = find_program('ratbagd', required: true)
+    ratbagd = find_program('ratbagd', required: false)
     ratbagd_required_version='0.17'
 
     if not ratbagd.found()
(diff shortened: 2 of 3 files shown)


staticssleever668:
Hi! flatpak file [in flathub repo]https://github.com/flathub/org.freedesktop.Piper) seems to have newer dependency versions, is there any reason they are older here?

staticssleever668:
Hm, also, what is the use case of in-tree flatpak file? I initially thought it allowed people to build Piper without having to install dependencies in the system, but it turns out it just clones this repo.

whot:
> Hi! flatpak file [in flathub repo](https://github.com/flathub/org.freedesktop.Piper) seems to have newer dependency versions, is there any reason they are older here?

oops. I thought I had done a rebase before copying over. Fixed now.

> what is the use case of in-tree flatpak file? 

originally it was so anyone could easily build a flatpak from whatever was in git at the moment - it predates the flathub approval and that one only does releases anyway. IIRC it was also designed so people can just `curl` the `.json` file and build a flatpak locally with whatever is on master right now.

We could change to `url: "."` in the flatpak file, that would install the current git tree. Though the argument is more that if you're developing, you don't really need a flatpak, right?

Or we could just drop the file and document the bits that are needed to adjust the flathub file. It doesn't seem to be used much anyway.

staticssleever668 (approved):

staticssleever668:
> Or we could just drop the file and document the bits that are needed to adjust the flathub file. It doesn't seem to be used much anyway.

After all, I think we should really just drop it.
One only use-case for it I thought of, is that it could be easier for non-dev people to contribute device artwork. However, I tried building the flatpak myself, and it turned out a bit more technical than I would like it to be. At this point, I think it would take about the same amount of effort as building Piper yourself.

staticssleever668:
By the way, I have less free time now as I've started working, and still haven't released a new version of libratbag with breaking API changes (and I feel guilty not finishing it for so long).
There is only one thing left I would like to change - it's adding modifiers to single key actions (also an API only just added).
What I've stuck on, is making D-Bus API for that. Maybe you could quickly help me sketch up how to send such data?
What I would like in theory, is sending a struct through D-Bus looking like so:
```c
struct KeyActionDbusData {
    u32 key;
    u32 modifiers[];
};
```
Frankly speaking, I was just being lazy and didn't play with D-Bus enough to make it work this way. Do you think that would be a good API? If so, is it possible to make it work this way? Just knowing if this is possible this way would help me finally push through this. My only other idea, which would actually be very easy to implement, is to just send an array with the key being on the first index, but that's not an elegant API.
(just in case, here is the PR where I was doing this: https://github.com/libratbag/libratbag/pull/1471)

whot:
> However, I tried building the flatpak myself, and it turned out a bit more technical than I would like it to be.

fwiw:

```
$ flatpak-builder -v --install --user --force-clean \
  --install-deps-from=flathub \
  builddir org.freedesktop.Piper.json
``` 

This should do the trick, at least that's the only command I've ever used. That installs it too so you can then `flatpak run org.freedesktop.Piper` or, since it installs the `.desktop` file it should then show up.

It's a little bit nicer than meson since it installs all deps so you don't have to go searching for them.


> What I would like in theory, is sending a struct through D-Bus looking like so:

For that struct [1] you would go with an `uau` signature (integer, followed by array of integers). Then you pack/unpack into that signature on both sides. We already have at least one API that uses `au` (`ratbagd_profile_get_capabilities`) so it's a matter of copying that code and prepending the key value first. roughly like this, I think:
```c
sd_bus_message_append(reply, 'u', key);
sd_bus_message_open_container(reply, 'a', "u");
for modifier in modifiers
   sd_bus_message_append(reply, 'u', modifier)
sd_bus_message_close_container(reply);
```


[1] fwiw structs in DBus have a slightly different meaning, so just because something is a struct in C doesn't mean it's a struct in DBus. Structs are (I think) more intended for inseparable elements, e.g. the RGB for our LEDs.

PR #894, opened 2023-08-30
Fix blank app icon issues with KDE application launcher

ahmetsait:
Removes all unused mask and clip paths, crops the embedded image, optimizes file size and fixes render issues seen in KDE apps (Gwenview, Application Launcher, Application Dashboard etc.).


staticssleever668:
Just curious, what tools have you used to make this change?

staticssleever668:
Image still shows up in Rofi and Piper's about page.
If you open up the new image in Inkscape, there are no groups that contain unrelated paths together anymore, everything is in the top level.
Curious how the shadow uses an embedded PNG instead of also being made in SVG. :thinking: 
LGTM, thank you!

staticssleever668 (approved):

ahmetsait:
> Just curious, what tools have you used to make this change?

Inkscape & Kate as text editor for fine grained edits.

> Image still shows up in Rofi and Piper's about page. If you open up the new image in Inkscape, there are no groups that contain unrelated paths together anymore, everything is in the top level. Curious how the shadow uses an embedded PNG instead of also being made in SVG. 🤔 LGTM, thank you!

This can be done better with an inner drop shadow filter or a mesh gradient but those are not well supported in anything other than browsers so I left that highlight effect as an image.

For reference this is how it looked in Gwenview before the changes:

image

PR #895, opened 2023-09-10
asus: add chakram x svg

kyokenn:


staticssleever668 (approved):
I checked it out in Piper, looks good!

PR #896, opened 2023-09-11
Add zh_CN translation

detiam:


staticssleever668 (approved):
I don't speak any Chinese languages, so I'll assume this is good

PR #931, opened 2024-01-18
add svg for asus rog harpe wireless

dleviminzi:
Adds link to default asus moouse for rog harpe


staticssleever668 (approved):
Thanks!
cc @kyokenn for ASUS

PR #957, opened 2024-05-17
Added svg glorious model o

DomHeadroom:


DomHeadroom:
Ok I will try to fix it later when I get home

DomHeadroom:
Ok I think the tool I used to reduce the size of the svg removed some metadata and now I need to get the original from my Desktop pc and I 'll be able to only in some months

PR #1025, opened 2025-03-07
Bump OARS version to 1.1

AFCMS:
Bumping OARS rating from 1.0 to 1.1 will remove unknown age rating categories from GNOME Software.

Currently it look like this:

image

It's a very very small contribution, basically sightly improving the app page on GNOME Software.


AFCMS:
Failed workflow doesn't seem to be related to the patch.

PR #1030, opened 2025-04-08
meson.build: make code formatting and linting optional

OriginCode:
Hi, as a Linux distribution packager, these options would be nice to be optional instead of mandatory as linting and code formatting are not necessary and may introduce unexpected modifications during build time.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,6 +13,8 @@ ratbagd_api_version = 2
 
 # Options
 enable_runtime_dependency_checks = get_option('runtime-dependency-checks')
+enable_formatting = get_option('format-code')
+enable_linting = get_option('lint-code')
 
 # Dependencies
 if enable_runtime_dependency_checks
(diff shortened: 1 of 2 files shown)


OriginCode:
> Hi! python-black-check and python-ruff-check tests only do the checking an don't modify any files. Still, I can understand why would they be unwanted. I don't understand `if enable_tests` though. If you don't want to run any tests then do just that - don't invoke `meson test`. Neither `meson compile` nor `meson install` run tests as far as I know.

Hi, I just removed the extra `enable_test` option.

staticssleever668:
1) These tests do not modify anything and are also skipped if `black` and `ruff` are not installed. My understanding is that distros usually build programs in a sandbox without unneeded libraries or programs installed.
2) I think these don't need to be two separate options, just one would suffice (even as `lint-code`)
3) Looks like Meson itself is going to add a flag soon to allow skipping specific tests https://github.com/mesonbuild/meson/issues/6999 https://github.com/mesonbuild/meson/pull/15739

eli-schwartz:
> These tests do not modify anything and are also skipped if `black` and `ruff` are not installed. My understanding is that distros usually build programs in a sandbox without unneeded libraries or programs installed.

Some do. Gentoo very specifically doesn't. Arch only does so optionally for official binary builds

> I don't understand `if enable_tests` though. If you don't want to run any tests then do just that - don't invoke `meson test`. Neither `meson compile` nor `meson install` run tests as far as I know.

This is perhaps more understandable to me -- the only other tests are appstreamcli validate and some scripts to verify svg source file consistency, so there isn't much for a distro to care about.

PR #1102, opened 2026-03-28
mouseperspective: display battery level via UPower

Rhorshaq:
ratbagd does not expose battery state in its DBus API. This commit adds a battery indicator (icon + percentage) in the MousePerspective header bar by querying UPower (org.freedesktop.UPower) independently.

A new UPowerDevice wrapper enumerates UPower mouse/gaming-input batteries and matches them to the ratbagd device by case-insensitive word-set intersection of their model names. If name matching fails and only one mouse battery is present, it is used unconditionally (common single-mouse setup). The widget updates live via the UPower PropertiesChanged signal and is hidden automatically for wired devices or when UPower is unavailable.

Known limitation: when multiple wireless mice are connected simultaneously and the name heuristic fails, no battery is shown rather than risking displaying the wrong device's battery level.

Assisted by Claude Sonnet 4.6 (claude-sonnet-4-6).

### DIFF
--- a/data/ui/MousePerspective.ui
+++ b/data/ui/MousePerspective.ui
@@ -193,6 +193,42 @@
         <property name="position">1</property>
       </packing>
     </child>
+    <child>
+      <object class="GtkBox" id="battery_box">
+        <property name="visible">False</property>
+        <property name="can-focus">False</property>
+        <property name="spacing">4</property>
+        <child>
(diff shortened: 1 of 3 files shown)


staticssleever668:
This is a neat idea, thank you, but libratbag is only meant as a configuration tool

PR #1106, opened 2026-04-23
Add G502 X Plus (Wireless bundled dongle ID c547)

renatofrota:
Add Logitech G502 X Plus wireless dongle ID c547 (I got this one, despite another dongle ID is known and is already supported)


staticssleever668:
c547 is a receiver so it could be any mouse, and c095 is already defined above

renatofrota:
c547 is a receiver that comes bundled with 502 X Plus (and has "G502 X" label on it) and is not present on the file yet, so I see no reason for this to be rejected.

albertorm95:
@staticssleever668 🔝

Issue #1104, opened 2026-04-12
Please add naming of profiles


jakethelizard99:
Instead of just having profiles 0-4, it would be nice to set up names for profiles, such as if using it for gaming purposes, you could add the label Battlefield or Dying Light, etc, or use Word or Excel for productivity.  I feel like this change would just make this application so much more functional than just having to remember what number profile is for what purpose, as it is now.

Pirateguybrush:
Seconded

PR #1109, opened 2026-05-14
feature: Add profile renaming

### LINKED ISSUE #85: Allow setting profile names
Not yet in libratbag, but once it does we should support it. See https://github.com/libratbag/libratbag/pull/275.

saaranganand:
Closes https://github.com/libratbag/piper/issues/85 and https://github.com/libratbag/piper/issues/1104

Adds the ability to rename profiles

a pencil button (document-edit-symbolic) appears next to each profile row on devices that support profile names. ratbagd has no capability flag for renaming, so a profile having a name is what indicates support.

Clicking it switches to an editable field. Enter / moving focus away chooses the new name. Esc cancels without saving. The header label updates as soon as teh profile is renamed.

Only works on older devices that use HID++ 1.0 (devices that support renaming): for example Logitech G5, G9/G9x, G500/G500s, G700/G700s, MX series, M705

Tested with the libratbag test device (rename persists in ratbagd). On devices without name support (tested with a G502 HERO), the button stays hidden

also fixed a possible crash where the `_on_properties_changed` handler could run on a partially initialized ratbagd if `__init__` raised (separate but minor issue in piper/ratbagd.py)

### DIFF
--- a/piper/profilerow.py
+++ b/piper/profilerow.py
@@ -10,7 +10,7 @@ from piper.ratbagd import RatbagdProfile
 from .util.gobject import connect_signal_with_weak_ref
 
 gi.require_version("Gtk", "3.0")
-from gi.repository import GObject, Gtk  # noqa
+from gi.repository import Gdk, GObject, Gtk  # noqa
 
 
 @Gtk.Template(resource_path="/org/freedesktop/Piper/ui/ProfileRow.ui")
@@ -21,10 +21,14 @@ class ProfileRow(Gtk.ListBoxRow):
     __gtype_name__ = "ProfileRow"
 
     title: Gtk.Label = Gtk.Template.Child()  # type: ignore
+    name_stack: Gtk.Stack = Gtk.Template.Child()  # type: ignore
+    name_entry: Gtk.Entry = Gtk.Template.Child()  # type: ignore
+    rename_button: Gtk.Button = Gtk.Template.Child()  # type: ignore
 
     def __init__(self, profile: RatbagdProfile, *args, **kwargs) -> None:
         Gtk.ListBoxRow.__init__(self, *args, **kwargs)
         self._profile = profile
+        self._committing = False
         connect_signal_with_weak_ref(
             self, self._profile, "notify::disabled", self._on_profile_notify_disabled
         )
@@ -34,6 +38,9 @@ class ProfileRow(Gtk.ListBoxRow):
             name = f"Profile {profile.index}"
 
         self.title.set_text(name)
+        self.rename_button.set_visible(
+            RatbagdProfile.CAP_WRITABLE_NAME in profile.capabilities
+        )
         self.show_all()
         self.set_visible(not profile.disabled)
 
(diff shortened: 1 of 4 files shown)


staticssleever668:
> also fixed a race condition where _on_properties_changed signal could happen before self._devices was initialized in Ratbagd.__init__ (separate but minor issue in piper/ratbagd.py)

How exactly can that happen?

Also, my mouse doesn't support profile names. Has anyone else tried this?

saaranganand:
> > also fixed a race condition where _on_properties_changed signal could happen before self._devices was initialized in Ratbagd.**init** (separate but minor issue in piper/ratbagd.py)
> 
> How exactly can that happen?

i suppose race condition isnt the right term. the signal handler is brought in at https://github.com/libratbag/piper/blob/0a1b2667aee0b2b7d40c561051911fc79f3a3317/piper/ratbagd.py#L282 

but `self._devices` isn't set until https://github.com/libratbag/piper/blob/0a1b2667aee0b2b7d40c561051911fc79f3a3317/piper/ratbagd.py#L290 

if either of the 2 errors (`RatbagdUnavailableError`/`RatbagdIncompatibleError`) are raised, the handler stays on the incomplete object and any later `Devices` signal crashes it
 
 i just initialized `self._devices = []` first to avoid this
 
 ---
 
> Also, my mouse doesn't support profile names. Has anyone else tried this?

the edit button would only show up if your mouse reports `CAP_WRITABLE_NAME`. that comes from libratbag. it's only set if the mouse's protocol actually supports storing a profile name onboard. Piper would just check the capability, so on your mouse the button probably stays hidden. so far i've only tested on a G502 HERO

piper/profilerow.py
```diff
@@ -50,6 +57,44 @@ def _on_delete_button_clicked(self, button: Gtk.Button) -> None:
...
+    def _on_name_entry_key_press_event(
+        self, entry: Gtk.Entry, event: Gdk.EventKey
+    ) -> bool:
+        if event.keyval == Gdk.KEY_Escape:
+            self.name_stack.set_visible_child_name("label")
+            return True
+        return False
+
+    def _commit_rename(self) -> None:
+        if self._committing:
```
staticssleever668: Is this to avoid some kind of recursion? Would be ideal to avoid this variable. If there's no way around it, it has to be renamed for clarity, something like '_is_in_rename_committing'.

staticssleever668:
I think this shows the rename button even on devices that don't support it.
        Even `self.rename_button.set_visible(False)` doesn't hide it.

piper/profilerow.py
saaranganand: yes i added that to stop it from running twice (hiding the text box after entering a name makes it lose focus, and losing focus calls this function again). but yeah we can just check at the start if the text box is no longer showing. will update

i dont think the check should be removed entirely, because of `esc`: pressing `esc` hides the text box too, and if that calls `focus-out` it would commit the typed name instead of cancelling

saaranganand:
> I think this shows the rename button even on devices that don't support it. Even `self.rename_button.set_visible(False)` doesn't hide it.

i believe thats because of this line:
https://github.com/libratbag/piper/blob/0a1b2667aee0b2b7d40c561051911fc79f3a3317/piper/profilerow.py#L37
it runs after the button is hidden and so re-shows everything. can probably add a property to the button ui itself so it stays hidden. will update

saaranganand:
> the edit button would only show up if your mouse reports CAP_WRITABLE_NAME. that comes from libratbag. it's only set if the mouse's protocol actually supports storing a profile name onboard. Piper would just check the capability, so on your mouse the button probably stays hidden. so far i've only tested on a G502 HERO

turns out `CAP_WRITABLE_NAME` doesn't exist in libratbag anymore. so yeah the button actually only showed on my G502 because of the `show_all()` bug above. ratbagd only accepts a name for a profile that already has one (currently just HID++ 1.0 devices, it seems). i've retested with the libratbag test device, and also my G502 now correctly shows no button

some HID++ 1.0 devices are Logitech G5, G9/G9x, G500/G500s, G700/G700s, MX series, M705

i'll update the PR description accordinlgy

PR #1120, opened 2026-07-21
Set prgname to application ID

City-busz:
Using the application ID ensures that Wayland compositors could match the window with the application and show the appropriate icon for them.

References:
- https://honk.sigxcpu.org/con/GTK__and_the_application_id.html
- https://docs.gtk.org/gtk4/migrating-3to4.html#set-a-proper-application-id


staticssleever668 (approved):
This fixes application icon on Wayland. Thanks!

PR #1124, opened 2026-07-31
Notify when the active profile changes

valsr:
Sends a desktop notification (via Gio.Notification) whenever the device's active profile changes, whether the switch was triggered from Piper or from the device itself (e.g. a hardware button). A checkable "Notify on Profile Switch" item in the primary menu, backed by a new org.freedesktop.Piper GSettings schema, lets users turn it off.


staticssleever668:
That's a neat idea, but it isn't really fit for Piper. For Piper specifically it's only important to show the correct active profile in the top-left. This notification functionality could live in a small program of it's own utilizing libratbag's D-Bus API.

PR #1129, opened 2026-08-12
window: don't read mouse_perspective.device unless it is visible

### LINKED ISSUE #1128: Piper crashes on startup with AssertionError in _on_device_removed
### Description

Piper dies about two seconds after launch with an `AssertionError` whenever ratbagd emits `device-removed` while the welcome perspective is showing, i.e. before any device has been selected. Nothing is ever drawn, so from the desktop it just looks like Piper never starts.

### Traceback

```
ago 12 12:17:38 systemd[4268]: Started Piper.
ago 12 12:17:38 systemd[1]: Started Daemon to introspect and modify configurable mice.
ago 12 12:17:40 piper[35431]: Traceback (most recent call last):
ago 12 12:17:40 piper[35431]:   File "/usr/lib/python3.14/site-packages/piper/window.py", line 123, in _on_device_removed
ago 12 12:17:40 piper[35431]:     if device is mouse_perspective.device:
ago 12 12:17:40 piper[35431]:                  ^^^^^^^^^^^^^^^^^^^^^^^^
ago 12 12:17:40 piper[35431]:   File "/usr/lib/python3.14/site-packages/gi/_propertyhelper.py", line 246, in __get__
ago 12 12:17:40 piper[35431]:     value = self.fget(instance)
ago 12 12:17:40 piper[35431]:   File "/usr/lib/python3.14/site-packages/piper/mouseperspective.py", line 68, in device
ago 12 12:17:40 piper[35431]:     assert self._device is not None
ago 12 12:17:40 piper[35431]:            ^^^^^^^^^^^^^^^^^^^^^^^^
ago 12 12:17:40 piper[35431]: AssertionError
```

### Cause

`Window._on_device_removed()` reads `mouse_perspective.device` unconditionally:

https://github.com/libratbag/piper/blob/master/piper/window.py#L119-L131

but `MousePerspective.device` asserts that a device has been set:

```python
@GObject.Property
def device(self) -> RatbagdDevice:
    assert self._device is not None
    return self._device
```

and `_device` is only assigned by `_present_mouse_perspective()`. When two or more devices are connected, `_on_device_added()` leaves Piper in the welcome perspective and `set_device()` is never called, so any `device-removed` signal arriving at that point raises.

The comment directly below the check already spells out the intended condition:

```python
# The current device disconnected, which can only happen from the
# mouse perspective as we'd otherwise be in the welcome screen with
# more than one device remaining.
```

but the code never actually checks which perspective is visible — unlike the `elif` branch immediately below it, which does exactly that with `get_visible_child_name() == "welcome_perspective"`.

### Reproducer

Any setup where ratbagd exposes more than one device and one of them goes away during ratbagd's cold start.

Here an ASUS ROG Keris (`usb:0b05:195c`) is enumerated twice, because the mouse exposes several HID interfaces and ratbagd enumerates every `hidraw` node without filtering by interface (`ratbagd/ratbagd.c`, `udev_enumerate_add_match_subsystem(e, "hidraw")`), while the `asus` driver does not check the interface number either:

```
$ ratbagctl list
warbling-mara:       ASUSTeK ROG KERIS
singing-gundi:       ASUSTeK ROG KERIS

$ ls -l /sys/class/hidraw/
hidraw0 -> .../3-3:1.0/0003:0B05:195C.0001/hidraw/hidraw0
hidraw1 -> .../3-3:1.1/0003:0B05:195C.0002/hidraw/hidraw1
hidraw2 -> .../3-3:1.2/0003:0B05:195C.0003/hidraw/hidraw2
```

Launching Piper from the desktop menu D-Bus-activates ratbagd (note the identical timestamps in the log above); ratbagd emits a `device-removed` during that cold start and Piper dies before its window is mapped. Launching Piper when ratbagd has already been running for a while usually works, which makes this present as an intermittent "Piper doesn't start" bug that is hard to pin down.

### Versions

- piper 0.8
- libratbag 0.18 (git, `0.18.r19.f9d15748`)
- Arch Linux (CachyOS), Python 3.14, GTK3, KDE Plasma / Wayland

### Fix

Checking the visible perspective before reading `.device` fixes it; with that change Piper survives ratbagd's cold start and shows the device chooser as intended. PR follows.

ayozetr:
Fixes #1128.

`Window._on_device_removed()` reads `mouse_perspective.device`, whose getter asserts `self._device is not None`. `_device` is only assigned in `_present_mouse_perspective()`, so while the welcome perspective is showing (two or more devices connected, none selected yet) a `device-removed` signal raises `AssertionError` and takes the application down before its window is ever mapped — it just looks like Piper does not start.

The check now looks at the visible perspective first, which is what the comment right below it already describes, and what the `elif` branch below it already does. No behaviour change once the mouse perspective is visible.

Tested on Arch with an ASUS ROG Keris that ratbagd enumerates twice (several HID interfaces, none filtered out). Before the change Piper died ~2s after launch whenever it was the process that D-Bus-activated ratbagd; after it, Piper survives ratbagd cold start and shows the device chooser as intended. `black` and `ruff` are clean.


staticssleever668 (approved):
This isn't the proper fix because it means there is some kind of race condition in entering 'mouse_perspective', but if it works, it's fine by me

ayozetr:
Thanks for the quick review and merge.

You're right that this only stops the crash rather than removing the race, so for the record, here is what it looked like on my side in case it helps whoever picks it up.

Piper is the process that D-Bus-activates ratbagd, so both start within the same second and the `AssertionError` lands about two seconds later:

```
12:17:38 systemd[4268]: Started Piper.
12:17:38 systemd[1]: Started Daemon to introspect and modify configurable mice.
12:17:40 piper[35431]: AssertionError
```

Launching Piper when ratbagd has already been running for a while does not reproduce it, which is why this presents as an intermittent "Piper doesn't start".

It only becomes fatal when ratbagd exposes the same device more than once. With a single device `_on_device_added()` calls `_present_mouse_perspective()` immediately, so `_device` is set and the old code was safe; with two, Piper stays on the welcome perspective with `_device` still `None` and any `device-removed` raises.

In my case the duplicate is libratbag/libratbag#1879 — the `asus` driver calls `ratbag_open_hidraw()` and so binds to every HID interface of the mouse, exposing one real configuration endpoint and one bogus entry that returns the floor of every range. Fixing that would remove the trigger for these devices.

What I could not pin down is *why* ratbagd emits `device-removed` during its cold start at all; I only have the evidence that it does, not the mechanism. That may well be worth a look on the libratbag side independently of the duplicate.

